PR #5, opened 2024-01-09
debug deploy

thedavidmeister:


thedavidmeister:
Need `--rpc-url` to be set directly https://github.com/foundry-rs/foundry/issues/6731

PR #37, opened 2024-03-07
build: add libusb, gmp

mattyg:
Resolves https://github.com/rainlanguage/rain.orderbook/issues/398

- confirmed it resolves the libusb issue when running sol-prelude
- did not confirm it resolves the gmp issue -- unable to get to that point due to other build issues with foundry-evm not pinning alloy


thedavidmeister:
this shouldn't be needed

PR #113, opened 2026-04-07
chore: Organize flake

eureka-cpu:
## Motivation

Just getting familiar with the code base. I see that the flake has a lot going on, and there's a good number of anti-patterns so this is really just a cleanup PR.

## Solution

Use `writeShellApplication` instead of custom wrapper type `mkTask`. `mkTask` is essentially trying to do what `writeShellApplication` already does, without the added benefit of running shellcheck.

Organize derivations and overlays for better composability, and less likelihood of getting or incorrectly setting policy decisions used by downstream consumers.

Add `legacyPackages` and `apps` for package introspection from the repl and discoverability.

Add the default package to `checks` for convenience.

## Checks

By submitting this for review, I'm confirming I've done the following:
- [ ] made this PR as small as possible
- [ ] unit-tested any new functionality
- [ ] linked any relevant issues or PRs
- [ ] included screenshots (if this involves a front-end change)

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -17,6 +17,21 @@
       }
     },
     "flake-utils": {
+      "locked": {
+        "lastModified": 1644229661,
+        "narHash": "sha256-1YdnJAsNy69bpcjuoKdOYQX0YxZBiCYZo4Twxerqv7k=",
+        "owner": "numtide",
+        "repo": "flake-utils",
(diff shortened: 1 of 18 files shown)


eureka-cpu:
If `mkTask` is used a lot downstream, to avoid breaking things we could change the function such that the inputs stay the same, but the actual returned value is `writeShellApplication` and using it prints a deprecation warning. That would make it a bit easier to eventually adopt the right pattern.

PR #217, opened 2026-06-01
lib: add update-rainix.sh helper for bumping consumers

thedavidmeister:
Adds `lib/update-rainix.sh`, a helper that bumps a rainix-consuming repo to the latest rainix and re-locks Soldeer. Run from a consumer repo root; it makes local changes only.

What it does:
- points the `rainix` flake input at `github:rainlanguage/rainix` and runs `nix flake lock --update-input rainix`
- normalises `flake.nix` for the current lint hooks — ensures the `outputs` argument set ends with `...` (so deadnix can drop an unused `self` without breaking evaluation), then runs deadnix/statix/nixfmt
- for Solidity repos, re-locks Soldeer (`forge soldeer install`) + sanity `forge build`

Soldeer dependency *version* bumps are intentionally left to the developer (edit `foundry.toml`, `forge soldeer update`, fix version-suffixed imports), since bumping blindly can break builds when a transitive dependency pins an older version with `recursive_deps = false`.

This is the distilled, in-place form of the script used to open the org-wide "latest rainix + soldeer re-lock" PRs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,10 +39,8 @@ commit yourself):
 /path/to/rainix/lib/update-rainix.sh
 ```
 
-It points the `rainix` flake input at `github:rainlanguage/rainix` and updates
-the lock, normalises `flake.nix` for the current lint hooks (ensures the
-`outputs` argument set ends with `...`, then runs deadnix/statix/nixfmt), and —
-for Solidity repos — re-locks Soldeer and runs a sanity `forge build`. Soldeer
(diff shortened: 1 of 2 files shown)


thedavidmeister:
Reviewed 24eec6750: minimal update-rainix.sh helper (bump rainix input + soldeer re-lock + sanity build); slimmed of migration hacks. No CodeRabbit findings. CI green. LGTM.

PR #221, opened 2026-06-06
feat(autopublish): soldeer auto-bumps + publishes on content change (like the crates)

thedavidmeister:
## What
Makes the shared `rainix-autopublish.yaml` **auto-bump + publish the Soldeer package on merge**, the same model as the crates — instead of the current version-divergence gate that requires a manual `foundry.toml [package].version` bump.

## How
Mirrors the cargo content-gate: hash the normalized content of `forge soldeer push --dry-run` vs the latest published revision's zip (registry download URL). On a content change → auto-bump `foundry.toml [package].version` (patch on the greater of local/registry, monotonic), commit (`Package Release` prefix so the push self-skips), tag `sol-v<next>`, publish at `<next>`. The version line is blanked in the hash so a bump alone never re-triggers (no infinite loop).

## Validation
Locally against raindex: published `0.1.0` hash `5d8ecf3` ≠ current-tree hash `7fd8a02` (correctly detects the 101 commits of drift since `sol-v0.1.0`); deterministic; `--dry-run` writes `<cwd>.zip` to cwd and works without an install. **Needs a CI run to confirm the end-to-end bump→tag→publish path** (the dry-run zip location + the bump commit + soldeer auth).

## Heads-up (intended) side effect
Once merged, the next push to `main` on every Soldeer consumer will detect content-drift (they're all behind their published version — e.g. raindex is 101 commits stale) and auto-bump+publish, catching the registry up to `main`.

### DIFF
--- a/.github/workflows/rainix-autopublish.yaml
+++ b/.github/workflows/rainix-autopublish.yaml
@@ -24,7 +24,7 @@ on:
         type: string
         default: ''
       soldeer-package:
-        description: optional Soldeer registry package name (e.g. rain-erc). When set, the workflow also publishes the current foundry.toml version to Soldeer when it diverges from the latest registry entry. Soldeer version is read from foundry.toml [package].version; no auto-bump (dev manages soldeer versions explicitly).
+        description: optional Soldeer registry package name (e.g. rain-erc). When set, the workflow content-gates the Soldeer package (dry-run zip vs the latest published revision) and AUTO-BUMPS foundry.toml [package].version on a content change, then publishes — same merge-driven model as the crates (no manual version bump needed).
         required: false
         type: string
         default: ''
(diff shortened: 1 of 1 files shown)


thedavidmeister:
Reviewed 4d8de3262: Change is isolated to the **soldeer path** — version-divergence gate replaced by a content gate, a bump step added, four `outputs.local`→`outputs.next` refs updated; cargo + npm publish paths untouched (nil risk there). Content gate mirrors the proven cargo gate; validated locally against raindex (pub 0.1.0 hash `5d8ecf3` ≠ current `7fd8a02` → detects the 101-commit drift; deterministic; `--dry-run` writes `<cwd>.zip`, no install needed); foundry.toml version line blanked in the hash so a bump can't re-trigger. Caveat: rainix CI here only tests the flake fixtures — the autopublish workflow runs solely on consumer merges, so the end-to-end bump→tag→publish is first exercised by the next consumer push; failure modes (zip-not-found/gate error) surface as a failed autopublish job, recoverable + soldeer-isolated. Want eyes on the first consumer run.

PR #223, opened 2026-06-07
fix(autopublish): soldeer next-version + idempotent bump (stop no-op-failing reruns)

thedavidmeister:
Follow-up to #221. The first real raindex autopublish run surfaced two bugs:

1. **No-op re-bump.** `next` was computed from the **registry** (`remote`) only. Once `foundry.toml` got ahead of the registry — a prior run bumped + tagged `sol-v0.1.1` but its **publish failed** (separate issue, below) — every rerun set `next = remote+1 = local = 0.1.1`, so the `sed` was a no-op, `git commit` died `nothing to commit`, and publish was never re-attempted (registry stuck at 0.1.0). **Fix:** `next = patch(max(local, remote))`; a first publish (`remote=none`) uses `local` as-is.
2. **Bump commit fragility.** It failed on an empty stage and ran the repo's pre-commit hooks. **Fix:** only commit when foundry.toml changed; `--no-verify`.

**Not changed:** the publish step. The original publish failure was soldeer's interactive *sensitive-files* warning (`error during IO operation: not connected` in CI, because raindex packages `.env.example`). The fix for that is on the **consumer side** — keep sensitive files out of the package via `.soldeerignore` — **not** `--skip-warnings` (that warning exists to stop publishing secrets and should stay).

### DIFF
--- a/.github/workflows/rainix-autopublish.yaml
+++ b/.github/workflows/rainix-autopublish.yaml
@@ -203,8 +203,15 @@ jobs:
           if [ -z "$URL" ] || [ "$REMOTE" = "none" ]; then OLD="none"; else
             curl -fsSL "$URL" -o /tmp/soldeer_pub.zip; OLD=$(nh /tmp/soldeer_pub.zip); fi
           echo "soldeer gate: remote=$REMOTE local=$LOCAL OLD=$OLD NEW=$NEW"
-          BASE="$REMOTE"; [ "$BASE" = "none" ] && BASE="$LOCAL"
-          NEXT=$(python3 -c "p=[int(x) for x in (('$BASE'.split('.')+['0','0','0'])[:3])]; p[2]+=1; print('.'.join(map(str,p)))")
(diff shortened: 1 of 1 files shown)


thedavidmeister:
Reviewed 3e42e545a: Two soldeer-path fixes to the autopublish reusable (follow-up to #221), crate/npm paths untouched. (1) next = patch(max(local,remote)) instead of patch(remote) — so foundry.toml left ahead of the registry by a prior bumped-but-unpublished run yields next > local, not a no-op; first publish (remote=none) uses local as-is. (2) Bump commit only runs when foundry.toml actually changed, with --no-verify — no more 'nothing to commit' failures or pre-commit-hook gating. Both validated by tracing the real raindex failure (registry 0.1.0, foundry 0.1.1, no-op rerun). Publish step deliberately unchanged (no --skip-warnings; sensitive files handled consumer-side via .soldeerignore, see raindex#2679). rainix CI green (it tests the flake fixtures; the autopublish path is exercised on the next consumer merge). No blockers.

PR #224, opened 2026-06-09
soldeer gate: ignore pinned deploy-constant blocks in the content hash

thedavidmeister:
## Problem — an autopublish feedback loop

raindex pins each published soldeer version's deployed addresses back into its **own published source** (`LibRaindexDeploy` `*_DEPLOYED_{ADDRESS,CODEHASH}_<x>_<y>_<z>` constants); `testAllPublishedSoldeerTagsHaveAFullConstantSuite` requires them for every published tag. But pinning those constants **is itself a `src/` change**, so the content gate sees new content → publishes the next version → which now needs *its* constants pinned → … We hit this live: pinning `0.1.2` published `0.1.3`.

## Fix

The base64'd hash helper now strips pinned `*_DEPLOYED_{ADDRESS,CODEHASH}_<x>_<y>_<z>` blocks (+ their doc comments) from `.sol` files and collapses the blank-line gaps before hashing. So pinning a version's constants is **not** a content change. The **un-versioned** current constants (no `_x_y_z` suffix) are untouched, so a real bytecode change still flips the hash and publishes. No-op for repos without the pattern (pure libraries).

Validated locally against the real `LibRaindexDeploy.sol`:
- adding a `_0_1_3` block → **identical** normalized hash (loop broken)
- changing a current constant → **different** hash (real release still detected)
- embedded base64 decodes byte-for-byte to the helper; YAML lints

Follow-up: with this merged, raindex can pin `0.1.3`'s constants without publishing `0.1.4`.

### DIFF
--- a/.github/workflows/rainix-autopublish.yaml
+++ b/.github/workflows/rainix-autopublish.yaml
@@ -183,14 +183,22 @@ jobs:
       # re-bump). On a content change the next step bumps foundry.toml
       # [package].version (patch on the greater of local/registry, kept
       # monotonic) so Soldeer publishes on merge like the crates — no manual
-      # version bump. The hash helper is base64'd to avoid YAML heredoc fragility.
+      # version bump. The hash helper also strips pinned
+      # `*_DEPLOYED_{ADDRESS,CODEHASH}_<x>_<y>_<z>` deploy-constant blocks from
+      # .sol files before hashing: a repo that records a published version's
+      # deployed addresses back into its own source (e.g. raindex's
+      # LibRaindexDeploy) would otherwise self-trigger an endless
+      # bump -> pin-constants -> bump loop, since pinning the constants is itself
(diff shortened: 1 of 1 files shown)


thedavidmeister:
Reviewed 7fb29716a: Soldeer content-gate fix — the base64'd hash helper now strips pinned `*_DEPLOYED_{ADDRESS,CODEHASH}_<x>_<y>_<z>` blocks (+ doc comments) from .sol files and collapses the blank-line gaps before hashing, so pinning a published version's deploy constants is no longer seen as a content change (breaks the bump→pin→bump loop). Un-versioned current constants are untouched, so real bytecode changes still publish; no-op for repos without the pattern. Validated locally: adding a _0_1_3 block → identical normalized hash, a real change → different hash, embedded base64 decodes byte-for-byte to the helper, YAML lints. 13/13 green. No blockers.

PR #225, opened 2026-06-13
Add one-contract-per-.sol-file static check (rainix#214)

### LINKED ISSUE #214: Standard check: fail when a .sol file declares more than one contract
## Ask

Add a standard check to rainix (so all consuming rain repos inherit it via the reusable `rainix-sol-static` workflow / `rainix-sol-static` task) that **fails when a Solidity `.sol` file declares more than one contract**.

Rain's convention is one contract per file (named after the file), but it's only enforced by review and drifts — e.g. helper/mocks get declared inline in `.t.sol` test files, and some source/helper files accumulate multiple contracts. A mechanical check keeps it consistent without reviewer effort.

## Suggested scope

- Count top-level `contract` and `abstract contract` declarations per `.sol`; fail if `> 1`.
- Open question for maintainers: should `library` and `interface` count too (stricter "one top-level definition per file"), or only deployable/abstract `contract`s? Recommend starting with `contract` + `abstract contract` and deciding on libraries/interfaces separately.
- File-scope `error`/`struct`/`enum`/`constant` declarations alongside a single contract should remain allowed (they're not contracts).

## Implementation options

- A small shell/awk check added to the `rainix-sol-static` task body (greps each tracked `.sol` for `^(abstract )?contract ` count), or
- A `forge lint` / slither custom check if one fits cleanly, or
- A pre-commit hook in the rainix hook bundle.

## Rollout caveat

This will fail CI in consumers that currently have multi-contract files (including some test helpers, e.g. files bundling a mock asset + a manager). Land it behind a cleanup pass, or stage it as a warning first, so it doesn't break downstream `main` on merge. Track the rain.vats cleanup at rainlanguage/rain.vats#314.

## Context

Surfaced while extracting inline `.t.sol` test helpers into dedicated `test/concrete/` files for the rain.vats #309 fix; a repo-wide audit there found several pre-existing inline-helper violations (rainlanguage/rain.vats#314). A rainix-level check would prevent this class of drift everywhere.

thedavidmeister:
## What

Adds a standard "one contract per `.sol` file" check that all consuming rain repos inherit via the reusable `rainix-sol-static` workflow.

- **`lib/sol-single-contract.sh`** — counts top-level `contract` / `abstract contract` declarations per file (ignoring matches inside `//` line comments and `/* */` block comments) and fails when any tracked `.sol` declares more than one. Per the issue's recommended scope:
  - `library` and `interface` are **not** counted.
  - File-scope `error` / `struct` / `enum` / `constant` declarations alongside a single contract remain allowed (they aren't contracts).
- **`rainix-sol-single-contract` task** (`flake.nix`) — runs the check over `git ls-files '*.sol'`. Added to `sol-tasks` so it is on `PATH` in `sol-shell`, and exported as a package so consumers can `nix run ..#rainix-sol-single-contract` locally.
- **`rainix-sol-static.yaml`** — invokes the task as a CI step alongside `slither .` and `forge fmt --check`, so every consumer of the reusable workflow inherits the check.
- **bats coverage** (`test/bats/task/sol-single-contract.test.bats`, wired into `default-shell-test`) — single/abstract/two-contract counts, library/interface not counted, line- and block-comment stripping, file-scope declarations allowed, and the git-tracked enumeration pass/fail paths.

## Why

Rain's convention is one contract per file, but it was only enforced by review and drifted — inline helper/mock contracts in `.t.sol` files and accumulated source helpers. A mechanical rainix-level check prevents this class of drift everywhere without reviewer effort.

## Rollout caveat (from the issue)

As the issue notes, this will fail CI in consumers that currently have multi-contract files (e.g. some test helpers bundling a mock asset + a manager). This PR lands the check in rainix; downstream consumers should be cleaned up before/as they pick it up (tracked for rain.vats at rainlanguage/rain.vats#314). The check is on tracked `.sol` files only and is a hard failure, matching the issue's primary recommendation; staging it as a warning first was an option but was not requested as the default.

The two open design questions in the issue (should `library`/`interface` count; warning-first vs hard-fail) are resolved conservatively here per the issue's own recommendation: start with `contract` + `abstract contract` only, hard-fail. They can be tightened separately if maintainers prefer.

## Verification

- `nix build .#rainix-sol-single-contract` and `nix build .#sol-shell-test` build green.
- The built task exits `0` on rainix's own tracked `.sol` fixtures and exits `1` with a clear message (`ERROR: <file> declares N contracts; ...`) on a tracked multi-contract file.
- All 12 bats tests pass.
- `shellcheck` (lib script), `nixfmt` (flake.nix), `yamlfmt` (workflow), and `deno fmt` (CLAUDE.md) are clean on the changed files.

Fixes #214

### DIFF
--- a/.github/workflows/rainix-sol-static.yaml
+++ b/.github/workflows/rainix-sol-static.yaml
@@ -46,3 +46,5 @@ jobs:
         run: nix develop github:rainlanguage/rainix#sol-shell -c forge soldeer install
       - run: nix develop github:rainlanguage/rainix#sol-shell -c slither .
       - run: nix develop github:rainlanguage/rainix#sol-shell -c forge fmt --check
+      # Enforce Rain's one-contract-per-.sol-file convention (rainix#214).
+      - run: nix develop github:rainlanguage/rainix#sol-shell -c rainix-sol-single-contract
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -35,6 +35,9 @@ All tasks are Nix packages run via `nix run`. From a consuming repo use `..#`
 - `nix run ..#rainix-sol-prelude` — `forge install && forge build`
 - `nix run ..#rainix-sol-test` — `forge test -vvv`
 - `nix run ..#rainix-sol-static` — `slither . && forge fmt --check`
+- `nix run ..#rainix-sol-single-contract` — fail if any tracked `.sol` declares
+  more than one top-level `contract`/`abstract contract` (one-contract-per-file
+  convention; `library`/`interface` not counted)
 - `nix run ..#rainix-sol-legal` — `reuse lint` (REUSE/DCL-1.0 license
   compliance)
 - `nix run ..#rainix-sol-artifacts` — deploy to testnet via `script/Deploy.sol`
--- a/flake.nix
+++ b/flake.nix
@@ -252,6 +252,20 @@
           additionalBuildInputs = sol-build-inputs;
         };
 
+        # Enforces Rain's one-contract-per-file convention (rainix#214). Runs
+        # over git-tracked .sol files so it works the same locally and in CI.
+        # `library`/`interface` are not counted; only `contract` and
+        # `abstract contract` declarations.
+        rainix-sol-single-contract = mkTask {
+          name = "rainix-sol-single-contract";
+          body = ''
+            set -euo pipefail
+            source ${./lib/sol-single-contract.sh}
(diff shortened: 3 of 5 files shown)


thedavidmeister:
Reviewed a969ac6d8: one-contract-per-.sol static check; fixed the reuse-lint false positive (an SPDX line inside a heredoc fixture in the new .bats file). All checks green. LGTM.

PR #226, opened 2026-06-14
ci: pin all third-party actions in reusable workflows to commit SHAs

thedavidmeister:
## What

Pin every third-party GitHub Action referenced by a **mutable tag** (`@v6`, `@v15`, etc.) in rainix's reusable workflows to the **immutable commit SHA** that tag currently points to. The exact current version is kept as a trailing `# <tag>` comment.

This pins the **exact current versions** — there is **no behavior change**, it is pure immutability hardening.

## Why

rainix owns the org's shared CI. These reusable workflows (`rainix-rs-test`, `rainix-sol-test`, `rainix-vercel`, `rainix-autopublish`, etc.) are inherited by **every consumer** — raindex, flow, and the rest. Referencing third-party actions by mutable tag is a supply-chain hole: a compromised or silently-retagged action would execute in every consumer's CI with their tokens and secrets. Pinning to a SHA makes the inherited CI **immutable and reproducible**.

It also **stabilizes the intermittent `cachix/cachix-action@v15` flake** currently hitting raindex CI (rs-static / wasm failing in the cachix step), since the action now resolves to a fixed commit instead of whatever `@v15` points to at job time.

This is the **upstream/source equivalent of raindex#2724** (the consumer-side equivalent). With rainix's reusables hardened here, raindex#2724 can be slimmed down to just its 2 repo-local pins.

## Scope

**70 references pinned** across **14 workflow files** / **11 distinct actions**:

| action | tag | pinned SHA | refs |
| --- | --- | --- | --- |
| actions/checkout | v6 | `df4cb1c069e1874edd31b4311f1884172cec0e10` | 13 |
| actions/checkout | v4 | `34e114876b0b11c390a56381ad16ebd13914f8d5` | 1 |
| actions/cache | v4 | `0057852bfaa89a56745cba8c7296529d2fc39830` | 5 |
| nixbuild/nix-quick-install-action | v30 | `5bb6a3b3abe66fd09bbf250dce8ada94f856a703` | 14 |
| cachix/cachix-action | v15 | `ad2ddac53f961de1989924296a1f236fcfbaa4fc` | 14 |
| nix-community/cache-nix-action | v7 | `7df957e333c1e5da7721f60227dbba6d06080569` | 13 |
| nix-community/cache-nix-action | v6 | `135667ec418502fa5a3598af6fb9eb733888ce6a` | 1 |
| Swatinem/rust-cache | v2 | `e18b497796c12c097a38f9edb9d0641fb99eee32` | 5 |
| jlumbroso/free-disk-space | v1.3.1 | `54081f138730dfa15788a46383842cd2f914a1be` | 1 |
| JS-DevTools/npm-publish | v3 | `19c28f1ef146469e409470805ea4279d47c3d35c` | 1 |
| softprops/action-gh-release | v2 | `3bb12739c298aeb8a4eeaf626c5b8d85266b0e65` | 2 |

`rainlanguage/*` reusable-workflow references are intentionally left on `@main` (first-party, not a third-party supply-chain concern). Anything already SHA-pinned was untouched (there was none).

## Notes

- Maintainers and **dependabot can still bump** versions via the `# <tag>` comments — that's the standard convention and keeps update tooling working.
- Annotated tags were dereferenced to their commit (`refs/tags/<tag>^{}`), not the tag object.
- All resolved SHAs were spot-checked against the GitHub commits API; all changed files remain valid YAML.

### DIFF
--- a/.github/actions/cache/action.yml
+++ b/.github/actions/cache/action.yml
@@ -1,0 +1,28 @@
+name: cache
+description: >-
+  Single-source wrapper around `actions/cache`. GitHub Actions can't
+  parameterize a `uses:` ref, so wrapping the step in a composite is the only
+  way to pin the action's SHA in ONE place across every rainix workflow that
+  caches build outputs. Each call site passes its own `path` / `key` /
+  `restore-keys` (the Foundry build cache, the npm cache, etc.).
+inputs:
+  path:
+    description: A list of files, directories, and patterns to cache and restore.
+    required: true
+  key:
+    description: An explicit key for restoring and saving the cache.
+    required: true
+  restore-keys:
+    description: >-
+      An ordered multiline string listing the prefix-matched keys used to
+      restore a stale cache if no cache hit occurred for `key`.
+    required: false
+    default: ''
+runs:
+  using: composite
+  steps:
+    - uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4
+      with:
+        path: ${{ inputs.path }}
+        key: ${{ inputs.key }}
+        restore-keys: ${{ inputs.restore-keys }}
--- a/.github/actions/checkout/action.yml
+++ b/.github/actions/checkout/action.yml
@@ -1,0 +1,22 @@
+name: checkout
+description: >-
+  Single-source wrapper around `actions/checkout`. GitHub Actions can't
+  parameterize a `uses:` ref (no `uses: ${{ env.X }}`), so the only way to pin
+  the checkout SHA in ONE place across every rainix workflow is to wrap it in a
+  composite. The `ssh-key` input covers the one call site (autopublish) that
+  needs a deploy-key checkout; every other site uses the default token checkout.
+inputs:
+  ssh-key:
+    description: >-
+      Optional deploy key (e.g. `secrets.PUBLISH_PRIVATE_KEY`) for a checkout
+      whose pushes should trigger downstream workflows. A composite action
+      cannot read `secrets.*`, so the caller must plumb the secret through here.
+      Empty (the default) falls back to the standard GITHUB_TOKEN checkout.
+    required: false
+    default: ''
+runs:
+  using: composite
+  steps:
+    - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6
+      with:
+        ssh-key: ${{ inputs.ssh-key }}
--- a/.github/actions/gh-release/action.yml
+++ b/.github/actions/gh-release/action.yml
@@ -1,0 +1,36 @@
+name: gh-release
+description: >-
+  Single-source wrapper around `softprops/action-gh-release`. GitHub Actions
+  can't parameterize a `uses:` ref, so wrapping the step in a composite is the
+  only way to pin the action's SHA in ONE place across the (npm / soldeer)
+  GitHub Release steps in the autopublish workflow. Each call site passes its
+  own `tag-name` / `name` / `files`.
+inputs:
+  tag-name:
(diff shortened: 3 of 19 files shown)


thedavidmeister:
## Self-test jobs are red: root cause + the bootstrap tradeoff

### Root cause (log evidence)
Every `Rainix CI` / `Rainix CI check shell` job dies in ~10s during **Prepare all required actions**:

```
Run ./.github/actions/nix-cachix-setup
Prepare all required actions
Getting action download info
Download action repository 'rainlanguage/rainix@main' (SHA:94a674a…)
##[error]Could not find file '…/rainix-94a674a…/LICENSE'.
```

The self-tests reference the new composites as `rainlanguage/rainix/.github/actions/<name>@main`. `@main` resolves to the **tip of main**, which does not yet contain `.github/actions/*` (they only exist on this branch), so the action download/staging fails. This is a pure **bootstrap** problem: a PR that first introduces same-repo composites referenced at `@main` cannot resolve those refs until the PR is *on* main.

### Why it can't be worked around on-branch without giving up an invariant
I tried the obvious fixes and verified against live CI / GH docs:

- **`./` refs** — work for a *workflow*, but inside a composite `./` resolves against the **caller's** repo (the downstream consumer), where `.github/actions/checkout` doesn't exist. Wrong for consumers. ([community#74934](https://github.com/orgs/community/discussions/74934), [runner#2431](https://github.com/actions/runner/discussions/2431))
- **`checkout: 'false'` to dodge `nix-cachix-setup`'s internal `checkout@main`** — doesn't help: GitHub resolves and **downloads every transitively-referenced action up front**; `if:` only gates *execution*, not the download. I pushed this and CI still failed at the same `checkout@main` download.
- **`./` in the self-test + a bootstrap `actions/checkout`** — goes green, but re-introduces the `actions/checkout` SHA outside the `checkout` composite (DRY count → 3), undoing the single-sourcing this PR exists to add.
- **Branch pin (`@<branch>`)** — green now, but dangles the instant the branch is deleted post-merge.

So `{green pre-merge, DRY count-1, zero post-merge skew}` is an **impossible trinity** for this bootstrap PR — you can have any two.

### What I did (and why)
Kept every ref at `@main` (so **DRY is intact — each third-party action single-sourced, count 1 — and there is zero version skew**) and added comments documenting the bootstrap. **The self-test jobs go green the moment this PR lands on main**: the failure is purely action *resolution* (`.github/actions/*` absent at `@main`); once merged they exist at `@main`, resolution succeeds, and the steps themselves are byte-equivalent wrappers of the same pinned actions that were already green. This is the option the issue prefers ("no post-merge skew"); its only cost is that the PR's own checks stay red until merge.

### If you'd rather see green *before* merge
Say the word and I'll push one of:
- **(A) inline `actions/checkout` inside `nix-cachix-setup`** (drop the internal composite ref) + `./` self-tests with a bootstrap checkout → green now, zero skew, but the `actions/checkout` SHA then lives in 2 places (partially undoes the DRY for checkout); or
- **(B) pin the self-test + internal-checkout refs to this branch's immutable commit SHA** → green now and never dangles, but the composites are then exercised at a frozen historical commit and want a one-line retarget to `@main` post-merge.

My recommendation is to **merge as-is** (zero skew, DRY fully intact, green on main) unless a green pre-merge check is a hard gate — in which case (A)/(B) trade one invariant for it. Your call.

thedavidmeister:
Reviewed 6a45dc77e: option A — merge as-is at @main. Every failing check is the composite-bootstrap (the reusable self-test jobs die at action-resolution because the single-sourced composites are referenced @main but only exist on this branch; they resolve + green once this lands on main). DRY invariant intact (each third-party action count 1), zero version skew. Will verify main green post-merge. LGTM.

PR #228, opened 2026-06-14
flake: keep sol-build-inputs chromium-free for sol-only shell (#128)

### LINKED ISSUE #128: Don't pull chromium into dev shell for Solidity-only repos
## Don't pull chromium into the dev shell for sol-only repos

The current `nix develop` shell pulls a chromium derivation. For repos that only need the Solidity toolchain (no JS/Rust/wasm), chromium is dead weight: large download, large nix store footprint, slow first-time bootstrap, and even with cache hits the closure adds substantial bytes that have to be extracted/linked.

## Where it bites

- First-time CI runs on a cold cache spend a long time fetching chromium (and its transitive deps).
- Local dev — every contributor on a clean machine pays the same one-time cost just to enter the shell, even when they're only editing Solidity.
- Cache size — chromium and its closure dominate the rainix cache footprint by a wide margin.

## What I'd like

Split the rainix dev shell so sol-only repos can opt into a slimmer environment. Either:

1. A separate `rainix-sol-shell` (or attribute / overlay) that has `forge`, `slither`, `reuse`, `solc`, `nixpkgs` essentials, but no chromium / playwright / npm / cargo / wasm-bindgen.
2. Make chromium an opt-in via a flake input or feature flag, rather than the default.

## Repos that would benefit

Every Rain repo whose CI is `rainix-sol-{test,static,legal}` only — i.e. doesn't touch JS or Rust:

- rain.solmem
- rain.math.binary
- rain.math.saturating
- rain.deploy
- rain.datacontract
- rain.string
- rain.sol.codegen
- rain.math.fixedpoint

That's most of the rainlanguage Solidity ecosystem.

### LINKED ISSUE #170: Ubuntu CI hangs building pkgs.chromium from source
## Symptom

All Ubuntu jobs in Rainix CI hang at the first `nix develop --command …` step after the Nix store cache restore, for hours, until they hit the 6h job timeout. macOS jobs in the same matrix complete normally (~10 min).

Examples from the post-merge run of PR #166:

| Job | URL | Hang point |
|---|---|---|
| `cargo build --release` ubuntu | https://github.com/rainlanguage/rainix/actions/runs/25799792336/job/75786363240 | step 5 `nix develop ../.. --command forge soldeer install` |
| Solidity test ubuntu | https://github.com/rainlanguage/rainix/actions/runs/25799792336/job/75786363265 | step 5 same |
| `rainix-check-shell` ubuntu | https://github.com/rainlanguage/rainix/actions/runs/25799792383/job/75786363118 | step 6 `nix develop --command cargo release --version` |

Live UI log shows nix building `chromium-unwrapped` from source.

## Cause

The default devShell on Linux pulls `pkgs.chromium` via `flake.nix:148`:

```nix
exec ${pkgs.chromium}/bin/chromium "$@"
```

PR #166 bumped nixpkgs to `4ba039de0909446943c07e2b42bd2f0f4507072e`. At that nixpkgs revision, `cache.nixos.org` does not have a prebuilt chromium binary, so Nix builds it from source on the runner — a multi-hour compile that exceeds the 6h job timeout under load.

macOS escapes because the same flake wraps the system Chrome.app on Darwin and never asks Nix to build chromium.

Slim shells (`sol-shell`, `rust-shell`) intentionally omit chromium, so consumers that pin to those (e.g. via `github:rainlanguage/rainix#sol-shell` in their reusable workflow calls) are unaffected. Only the default shell — which the rainix repo's own CI exercises — is impacted.

## Short-term fix (this PR)

Drop `pkgs.chromium` from the Linux default devShell. The chromium binary is currently only used to satisfy the `chromium.test.bats` smoke test and any downstream consumer that opens a headless browser; nothing in rainix itself actually invokes it. Removing it unblocks CI immediately and is safe to revert once a longer-term fix is in place.

## Long-term fix

Pick one:

1. Pin nixpkgs to a revision where `cache.nixos.org` has a prebuilt chromium binary (verify before bumping by querying the cache for the chromium output path).
2. Replace `pkgs.chromium` with a prebuilt binary fetch (`fetchzip` of a stable Chromium tarball, or `pkgs.google-chrome` if cached).
3. Add a project binary cache (cachix or similar) that stores the chromium build, so subsequent jobs pull rather than rebuild.

Option 1 is the smallest delta and is the default until/unless we hit it again.

thedavidmeister:
## What

Issue #128 asks that the dev shell for Solidity-only repos not pull chromium.

On investigation, **this is already the case on `main`**:

- The slim `sol-shell` devShell already exists (option 1 from the issue): `forge`, `slither`, `solc 0.8.25`, `reuse`, `jq`, `git` — no chromium, no node, no rust, no wasm. Every sol-only CI workflow (`rainix-sol-test`/`-static`/`-legal`, `rainix-copy-artifacts`, autopublish dry-run, etc.) consumes `nix develop github:rainlanguage/rainix#sol-shell`.
- Chromium is in **no** rainix dev shell at all: commit 37c992d (Closes #170) removed `pkgs.chromium` from the `default` shell because building it from source on a cold cache was exceeding the 6h CI timeout.

Verified against the realized closure of `sol-shell` on `x86_64-linux`:

```
$ nix-store -q --requisites $(nix build --no-link --print-out-paths .#devShells.x86_64-linux.sol-shell) | grep chromium
# (no output) -> chromium ABSENT
```

`nix run .#sol-shell-test` passes, including the regression guards `chromium should NOT come from nix` (slim.test.bats) and `sol-shell closure has no chromium` (closure.test.bats). The `sol-shell` still provides the full sol toolchain (forge / solc / slither / reuse / jq); the wasm/browser-test shell still carries `wasm-pack`.

## Change

Since the slim shell already satisfies the ask, this PR makes the invariant durable at its source: a comment on `sol-build-inputs` (the list shared by `sol-shell` and `default`) recording that it is the chromium-free shared Solidity set sol-only repos depend on. This is the exact place a future browser/chromium dependency would leak into every sol-only consumer, so the contract is documented where edits happen. Flake-only, nixfmt-clean, derivation-identical (the `sol-shell` drv path is unchanged).

Closes #128.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -89,6 +89,13 @@
         ])
         ++ (pkgs.lib.optionals pkgs.stdenv.isDarwin [ pkgs.darwin.DarwinTools ]);
 
+        # The Solidity toolchain shared by sol-shell and default. sol-only
+        # repos (rain.solmem, rain.deploy, rain.datacontract, etc.) consume
+        # this set via `nix develop github:rainlanguage/rainix#sol-shell` and
+        # nothing else, so it carries no browser (chromium/playwright), node,
+        # rust, or wasm dependency — keeping the sol-shell closure slim. The
+        # sol-shell closure test asserts chromium, node, and the rust
+        # toolchain stay absent.
         sol-build-inputs = [
           pkgs.git
           pkgs.foundry-bin


thedavidmeister:
Reviewed 488139af4: approved by maintainer; all 13 checks green, mergeable, CodeRabbit clean. Merging.

PR #240, opened 2026-06-28
fix(ci): plumb github-token cross-org so the flake-fetch 429 fix works

thedavidmeister:
## Problem

#239 added `access-tokens = github.com=${{ github.token }}` to the shared `nix-cachix-setup` preamble, but the 429 **persisted for cross-org consumers** — S01-Issuer's st0x deploy 429'd again ~40 min after #239 merged. Root cause: `${{ github.token }}` evaluates to **empty inside a composite invoked across the org boundary**, so the access-token line was effectively `github.com=` (no token) and the rainix flake fetch stayed unauthenticated.

## Fix

Add a `github-token` input to the composite and use `access-tokens = github.com=${{ inputs.github-token || github.token }}` — the explicit input wins, bare `${{ github.token }}` is the same-org fallback. The **12 cross-org reusables** pass `github-token: ${{ github.token }}` (which **is** populated in the reusable workflow's own context, unlike the composite's); `check-shell.yml` + `test.yml` (rainix's own same-org CI) omit it and ride the fallback. Consumers need no change — they already get a run token automatically.

Builds on #239.

## Summary by CodeRabbit

* **New Features**
  * Added support for passing a GitHub token into the shared CI setup.
  * Updated multiple workflows to provide the built-in token during setup, improving reliability for GitHub API access in automated runs.

* **Bug Fixes**
  * Reduced the chance of rate-limit issues during dependency resolution in CI, especially across repository boundaries.

### DIFF
--- a/.github/actions/nix-cachix-setup/action.yml
+++ b/.github/actions/nix-cachix-setup/action.yml
@@ -26,6 +26,16 @@ inputs:
       Optional `gc-max-store-size-macos` for the bundled cache-nix-action. Left empty (the action's own default) unless a caller needs to cap the macOS store before saving.
     required: false
     default: ''
+  github-token:
+    description: >-
+      GitHub token used to authenticate nix's GitHub API calls (flake
+      resolution), so they use the run-token rate limit (~1000/hr) instead of
+      the unauthenticated 60/hr/IP cap that 429s under CI bursts. `github.token`
+      does NOT populate inside a composite invoked across the org boundary, so a
+      cross-org reusable caller MUST pass its own `${{ github.token }}` here;
+      same-org callers may omit it and fall back to the composite's own token.
+    required: false
+    default: ''
 runs:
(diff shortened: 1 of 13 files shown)


thedavidmeister:
Reviewed e9389583d6413e21e623c8f0f33b981257425e2f: cross-org follow-up to #239. Adds a `github-token` input to `nix-cachix-setup` and sets `access-tokens = github.com=${{ inputs.github-token || github.token }}` (explicit input wins for cross-org callers; bare `${{ github.token }}` is the same-org fallback), plumbing `github-token: ${{ github.token }}` through the 12 cross-org reusables. Fixes the persistent cross-org 429 — `github.token` is empty inside a composite invoked across the org boundary but populated in the reusable's own context. `check-shell.yml`/`test.yml` ride the fallback. CI-config only, builds on #239, low risk. Approve.

PR #243, opened 2026-06-28
chore(ci): temporary token-wiring debug (log lengths only)

thedavidmeister:
Temporary debug to diagnose the persistent cross-org flake-fetch 429 despite the github-token plumbing from #240/#242. Adds one step to nix-cachix-setup that logs the LENGTHS (never values) of inputs.github-token (what the plumbing delivers) and bare github.token (cross-org-empty hypothesis). Will be reverted once the wiring is confirmed.

PR #244, opened 2026-06-28
chore(ci): extend token-wiring debug (prefix + rate_limit + nix.conf)

thedavidmeister:
Follow-up to #243. The token is present (len=426) yet the flake fetch still 429s. This logs: token prefix (type — ghs_ vs JWT), a rate_limit auth check (does the token actually authenticate?), and whether access-tokens landed in /etc/nix/nix.conf (i.e. does nix develop even read it). Temporary; revert after diagnosing. NOT self-merging — for review/merge by a human.

### DIFF
--- a/.github/actions/nix-cachix-setup/action.yml
+++ b/.github/actions/nix-cachix-setup/action.yml
@@ -59,21 +59,23 @@ runs:
         nix_conf: |
           keep-env-derivations = true
           keep-outputs = true
-          # Authenticate nix GitHub API calls (flake resolution) so they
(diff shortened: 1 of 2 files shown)


thedavidmeister:
Superseded. Diagnosis complete: the 429 is GitHub's burst rate-limit on the flake HEAD-resolution call (api.github.com/commits/HEAD), not a token-wiring issue — the token is present and valid (ghs_, applied via NIX_CONFIG) but the fetch still fails. Real fix is pinning the flake rev: #245. Closing this debug PR.

PR #251, opened 2026-07-02
feat(ci): enforce pragma convention ^ lib/abstract, = concrete (closes #250)

### LINKED ISSUE #250: Static check: enforce pragma convention (^ for lib/abstract files, = for concrete) alongside single-contract rule
## Rule

Org pragma convention: **`pragma solidity ^X.Y.Z` for files whose top-level declarations are only `library` / `interface` / `abstract contract`; `pragma solidity =X.Y.Z` for files declaring a concrete `contract`** (concrete test mocks/utils included). Libs/abstracts float so downstream soldeer consumers can compile published packages' `src/`; concretes pin.

Today this is convention-by-discipline only, and it drifts: audit passes flag the `^`/`=` mix as "inconsistent" and produce mass-pinning PRs in the wrong direction (e.g. rain.erc4626.words#225, human-rejected for pinning 6 src lib/abstract files; its issue rain.erc4626.words#55 shows the confusion the unenforced rule causes).

## Proposal

Enforce it mechanically in rainix, as a sibling of the one-contract-per-file rule:

- `lib/sol-pragma-convention.sh` following the `lib/sol-single-contract.sh` pattern (comment-stripping + top-level declaration scan is already written there and can be shared/reused):
  - file declares a top-level non-abstract `contract` → pragma must match `^pragma solidity =`;
  - file declares only `library`/`interface`/`abstract contract` (or just file-scope errors/constants/types) → pragma must match `^pragma solidity \^`;
  - version number itself stays repo-defined — the check enforces the operator, not the version;
  - skip generated (`src/generated/`) and vendored dirs, matching the existing checks' scoping.
- `rainix-sol-pragma-convention` mkTask in `flake.nix`, wired into the `rainix-sol-static` reusable next to `rainix-sol-single-contract`.
- bats coverage mirroring `test/bats/task/sol-single-contract.test.bats`.

## Edge cases to decide during implementation

- Empty workaround contracts in error-definition files (e.g. rain.flare's `contract ErrFtso {}` for foundry-rs/foundry#6572 inside an otherwise file-scope-errors-only file currently at `^0.8.19`): exempt empty contracts, or accept `=` there?
- Files with range/`>=` pragmas or multiple pragma lines: reject outright.
- Repos that intentionally deviate need a per-repo opt-out consistent with how other rainix checks handle exceptions.

thedavidmeister:
Closes #250

## What

Adds `rainix-sol-pragma-convention`, a new `sol-shell`-available task and `rainix-sol-static` step that mechanically enforces Rain's pragma operator convention:

- **`pragma solidity ^X.Y.Z`** — required for files whose top-level declarations are only `library`, `interface`, or `abstract contract` (plus any file-scope errors/structs/constants/types). The `^` lets downstream soldeer consumers compile published `src/` against any compatible compiler.
- **`pragma solidity =X.Y.Z`** — required for files declaring any concrete (non-abstract) `contract`. The `=` pins the compiler version exactly, so deployed bytecode is reproducible.

Concrete test mocks and utils are counted as concrete contracts — same rule, no exception.

## Components

- **`lib/sol-pragma-convention.sh`** — shell library following the `sol-single-contract.sh` pattern; comment-stripping and concrete-contract counting (column-zero `contract ` keyword after stripping), plus a `sol_pragma_convention_check_tracked` entry point that skips `src/generated/` and `dependencies/`
- **`flake.nix`** — `rainix-sol-pragma-convention` mkTask exported in `packages` and included in `sol-tasks` (available in `sol-shell`); `default-shell-test` bats runner extended
- **`test/bats/task/sol-pragma-convention.test.bats`** — 27 bats tests covering `sol_expected_pragma_operator`, `sol_actual_pragma_operator`, `sol_pragma_convention_check`, and `sol_pragma_convention_check_tracked`; all pass locally
- **`.github/workflows/rainix-sol-static.yaml`** — new step running `rainix-sol-pragma-convention` after the existing single-contract step (pinned SHA; consumers pick it up on the next flake bump)

## Edge cases decided

- **Empty workaround contracts** (e.g. `contract ErrFtso {}` for foundry#6572 alongside file-scope errors): treated as concrete contracts and require `=`. Cleaner than an "empty body" exemption and a one-time migration for any repos using this pattern.
- **Range/`>=`/bare pragmas**: reported as an unsupported operator (neither `^` nor `=`).
- **Generated/vendored files**: `src/generated/` and `dependencies/` are excluded (matches existing checks' scoping).

## Test plan

- [x] 27/27 bats tests pass locally (`bats test/bats/task/sol-pragma-convention.test.bats`)
- [ ] `check-shell.yml` passes (runs `default-shell-test` which includes the new bats suite)
- [ ] `test.yml` passes (sol + rs tasks against `test/fixture/`)

Generated with [Claude Code](https://claude.com/claude-code)


thedavidmeister:
🤖 ai:vetter
Reviewed 13aeda32dbdba233b1d2d310b08a6230a3b031db: ready — Merge-main re-vet: sol-pragma-convention.sh unchanged and correct (comment strip, column-zero concre
cost 335 — org-wide CI gate shell logic

thedavidmeister:
Human reject (David-authorized 2026-07-10) — reimplement in **Rust**, not shell + bats. Per the org de-bash pattern (de-bashed logic ships as a flake package invoked DIRECTLY — `rainix-static` is the template, never a thin bash wrapper), the pragma-convention check should be:

- a Rust flake package (`rainix-sol-pragma-convention`) doing the comment-strip + concrete-contract pragma scan in Rust, on PATH, wired into `rainix-sol-static.yaml`;
- `cargo test` unit tests over the pure scan logic + fixtures, replacing the BATS suite.

Keep the exact behaviour this PR already nailed: `^` for lib/abstract, `=` for concrete (incl. concrete test mocks); skip `generated`/`dependencies`; and the decided edge cases (empty-workaround contract → `=`, bare/`>=` → unsupported). Closes #250 once reworked in Rust.

PR #264, opened 2026-07-10
feat(autopublish): soldeer next-version release lifecycle (one process)

thedavidmeister:
Replaces the soldeer autopublish behaviour with the **single next-version lifecycle** — no mode flag, no branch. `[package].version` is the NEXT, unpublished version; on a content change vs the published revision the job publishes THAT version, bumps to the next, runs an optional generate hook, and commits. One merge-driven process every consumer adopts (adoption = a one-time toml bump to the next unpublished version; the generate hook is empty for repos with nothing to generate).

## Changes (soldeer path only)
- **Source-only change gate.** The content hash now excludes `src/generated/` (per-release `<tag>/` snapshots + generated aliasing libs). They're derived from source and a fresh `<tag>/` dir appears every release, so hashing them would flag "changed" on every merge and republish identical bytecode forever. This also **removes the old per-constant `*_DEPLOYED_..._x_y_z` pin-stripping**: the versioned pins live in the excluded snapshots, so pinning can no longer self-trigger the bump→pin→bump loop that hack existed to prevent.
- **Publish the in-dev version.** Publishes `[package].version` from the **pre-bump** tree so the package contents match the version it's published under, then bumps.
- **New `soldeer-generate-cmd` input** (optional). Runs in the sol-shell right after the bump, before the release commit, to re-key per-release artifacts; `git add -A` stages its output into the born-green "Package Release" commit. Empty default = no-op.
- **Stale-toml guard.** If `[package].version` isn't ahead of the registry (e.g. a prior run published but its bump-commit push failed), the gate **fails loud** with a clear action (bump the version) rather than silently re-publishing / mis-bumping — matching the 'CI never red without a clear action to green it' philosophy.
- No `soldeer-next-version` toggle: there is one process, not two.

## ⚠️ Review-critical (can't be end-to-end tested from a branch)
Autopublish only exercises on a real push-to-main, so this is a **draft for review**. The parts that most warrant your eyes:
1. **Publish-before-push ordering** (publish → bump-commit → tag+push) and its stale-toml recovery on a partial failure.
2. The `git add -A` capture of the generate output into the release commit.
3. The interaction with the shared `Tag and push` rebase step.

Suggest a **sandbox-repo test-merge** before this lands.

## Paired change
st0x.deploy #243 (the first next-version consumer): drop `soldeer-next-version`, set `soldeer-generate-cmd` to the forge `BuildPointers` regen. It's the reason this input existed as a not-yet-supported call, and it merges after this.

## Summary by CodeRabbit

* **New Features**
  * Enhanced automated Soldeer package releases with a “next-version” lifecycle, deterministic content gating, and consistent version-based release/tagging.
  * Added an optional post-bump generate command step during release automation.
  * Extended Rust-based Rainix tooling for CI checks and dev-shell availability.
* **Bug Fixes**
  * Improved reliability of content-change detection and enforcement that local versions advance beyond the latest published revision.
* **Documentation**
  * Updated tooling guidance to keep substantive logic in Rust and limit CI shell usage to orchestration.


thedavidmeister:
🤖 ai:vetter
Reviewed 0c2a9232a1bd664e3e9b8daf5a04c81440d15967: design — implements st0x.deploy#244's next-version lifecycle (paired with st0x.deploy#243) — Rust soldeer-gate + workflow rewiring are locally sound (fail-closed version parse, publish-before-bump, loud stale-toml guard), but this is an org-wide release-process decision: every soldeer consumer must one-time bump foundry.toml or its next content merge fails the gate; and rainix-static/curl enter sol-shell only in this PR, so the gate is command-not-found until the post-merge RAINIX_SHA bump (sequencing: land flake change + sha bump before the workflow switch, or sandbox test-merge as the body itself requests) — human ruling on the lifecycle before this lands
cost 812 — org-wide publish lifecycle, rust gate, bootstrap ordering

thedavidmeister:
Reviewed deb7a4f67: Soldeer content gate ported from inline bash (jq/awk/sha256sum/sort -V) to a Rust `rainix-static soldeer-gate` subcommand; rainix-static broadened to general rainix tooling on every shell's PATH (via common-shell-inputs), curl added to sol-build-inputs, and main.rs split into no_submodules.rs + soldeer_gate.rs modules. All gate logic (foundry.toml read, registry JSON parse, next-version ahead-invariant, normalize-hash with src/generated exclusion + version blanking, semver bump) is pure Rust with 14 unit tests running in the flake build's doCheck. Validated locally: nix flake check --impure, nix build .#rainix-static, CI-way sol-shell closure test (no rust-toolchain leak), reuse lint. CI green on ubuntu + macos. Maintainer approved merge ("merge it"). Follow-up after merge: bump RAINIX_SHA to the merge commit so the pinned sol-shell carries curl + the binary.

PR #268, opened 2026-07-11
feat(rainix-static): enforce append-only deploy-pin snapshots in CI

thedavidmeister:
## What

Enforce, in CI, that per-tag deploy-pin snapshots are **append-only**.

- New `rainix-static snapshots-append-only [--base <ref>] [--root <dir>]` subcommand: diffs the branch against the base and fails on any **modified or deleted** file under an existing `<root>/<tag>/` dir (tag = `<n>_<n>_<n>`, e.g. `0_1_4`). Adding a new `<tag>` is fine; non-tag files directly under `src/generated/` (the mutable current-pin deploy libs) are ignored.
- New `frozen-snapshots-append-only` composite action, wired into `rainix-sol-static`. No-op for repos without such snapshots.

## Why

A per-tag snapshot (`src/generated/<tag>/*.pointers.sol`) records a released contract's deterministic address + codehash. Downstream consumers pin those constants, so a snapshot must **never change on main or diverge between branches** — a release ships new bytecode by **adding a new tag** (bumping `[package].version`), never by rewriting an existing snapshot. This turns that immutability from a convention into a CI gate.

Companion to #267 (which stops the autopublish from pre-generating a next-version snapshot at bump). Together: snapshots are created by the PR that defines a version, and can never be silently changed afterward.

## Validation

- 18 unit tests (`is_tag`, `is_snapshot`, offender parsing — new tag vs modify/delete).
- End-to-end against a real violation: a branch that rewrites an existing `0_1_4` snapshot **fails** with a clear message; a clean `main` **passes**.

## Note

The shared checkout is shallow, so the composite unshallows before the three-dot (merge-base) diff — cheap for rain-sized repos. Base is `GITHUB_BASE_REF` on PRs, else `main`.

### DIFF
--- a/.github/actions/frozen-snapshots-append-only/action.yml
+++ b/.github/actions/frozen-snapshots-append-only/action.yml
@@ -1,0 +1,29 @@
+name: frozen-snapshots-append-only
+description: >-
+  Fails when a branch modifies or deletes an existing per-tag deploy-pin snapshot under src/generated/<tag>/ (tag = three underscore-separated numbers, e.g. 0_1_4). These snapshots are frozen once on the base branch — downstream consumers pin the address/codehash constants they hold, so a release ADDS a new tag (bump the version), never edits an existing snapshot. No-op for repos with no such snapshots.
+runs:
+  using: composite
+  steps:
+    - name: Enforce append-only snapshots
+      shell: bash
+      run: |
+        set -euo pipefail
+        # Nothing to enforce if the repo has no per-tag snapshots.
+        if ! ls src/generated/*/*.pointers.sol >/dev/null 2>&1; then
+          echo "snapshots-append-only: no per-tag snapshots; skip"
+          exit 0
+        fi
+        # The three-dot diff needs the merge-base with the base branch, but the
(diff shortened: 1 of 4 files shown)


thedavidmeister:
🤖 ai:vetter
Reviewed 5ddc1fcadf5f2aa16a268d8ffdf21d4dabff17f3: ready — companion to #267 (no linked issue) — rainix-static snapshots-append-only gate: is_tag/is_snapshot/parse_offenders are discriminating unit tests (new-tag allowed vs modify/delete flagged), three-dot merge-base diff --no-renames --diff-filter=MD is correct, composite action no-ops when no per-tag snapshots exist; logic in Rust + tested per repo doctrine
cost 332 — new rust subcommand + composite action, tests+git-diff read

thedavidmeister:
@coderabbitai review

thedavidmeister:
Reviewed 5ddc1fcad: adds rainix-static snapshots-append-only + the frozen-snapshots-append-only composite wired into rainix-sol-static, failing any branch that modifies/deletes an existing per-tag snapshot. 18 unit tests + verified end-to-end against a real violation; all CI green. Companion to #267. Approved for merge by thedavidmeister.

PR #272, opened 2026-07-15
fix: require script/Build.sol in copy-artifacts, no silent skip

thedavidmeister:
`rainix-copy-artifacts` gates regeneration on an exact filename:

```yaml
- name: Regenerate pointer artifacts
  if: hashFiles('script/BuildPointers.sol') != ''
  run: ... forge script ./script/BuildPointers.sol
```

A repo whose codegen script is named anything else doesn't fail — the step is **skipped**, then `forge build` + `forge fmt` + `git diff` run and pass. The job goes green having regenerated nothing, so the committed artifacts are no longer currency-checked at all. A stale-detection check that silently stops detecting is worse than not having one, because the green is read as proof.

Observable now on https://github.com/rainlanguage/rain.sol.codegen/pull/31, which renames its script:

```
skipped    Regenerate pointer artifacts
success    Build Solidity
success    Format
failure    Assert committed artifacts match freshly built   <- only a fmt diff
```

## Change

The codegen script is `script/Build.sol`, matched exactly. If `src/generated/` is committed and `script/Build.sol` is absent, the job **errors**. No legacy name, no `hashFiles` skip — a repo that renames or drops its script goes red instead of quietly reporting green over an unchecked tree.

Tolerating the legacy name was the obvious migration path and is deliberately not taken: a second accepted name keeps the hazard alive (the guard still can't tell "renamed" from "absent") and lets the rename defer forever. Repos that don't want to upgrade on this workflow's schedule pin rainix to a SHA.

## This is a flag day

On merge, every repo below goes red until it renames `script/BuildPointers.sol` -> `script/Build.sol`:

| repo | `src/generated` | regen script |
|---|---|---|
| rain.erc4626.words, rain.flare, rain.math.float, rain.merkle, rain.pyth, rain.sol.codegen, raindex, rainlang | yes | `BuildPointers.sol` -> **must rename** |
| rain.metadata | no | none — unaffected |

The fix per repo is `git mv script/BuildPointers.sol script/Build.sol`. `rain.metadata` has nothing generated, so the new guard never fires for it.

## Related

- https://github.com/rainlanguage/rain.sol.codegen/pull/31 — renames codegen's own script and drops the `.pointers.sol` suffix from `LibFs`/`LibSnapshot`. Must not merge before this lands, or its artifact check passes while skipping regeneration.

### DIFF
--- a/.github/workflows/rainix-copy-artifacts.yaml
+++ b/.github/workflows/rainix-copy-artifacts.yaml
@@ -29,32 +29,19 @@ jobs:
       - name: Regenerate meta artifacts
         if: hashFiles('script/build-meta.sh') != ''
         run: nix develop github:rainlanguage/rainix/${{ env.RAINIX_SHA }}#sol-shell -c ./script/build-meta.sh
-      # A repo that commits generated sources but exposes no regen step passes
-      # every check below without anything being regenerated, which is
-      # indistinguishable from a real pass. Fail loudly instead.
-      - name: Assert committed generated sources have a regen step
+      # Committed generated sources must be regenerable here, or the currency
+      # check below passes without checking anything. The codegen script is
+      # `script/Build.sol`, matched exactly: a repo that renames or drops it
+      # goes red rather than skipping regeneration and reporting green.
+      - name: Regenerate generated sources
         run: |
-          if [ -d src/generated ] \
-            && [ ! -f script/Build.sol ] \
-            && [ ! -f script/BuildPointers.sol ] \
-            && [ ! -f script/build.sh ]; then
-            echo "::error::src/generated/ is committed but no regen step was found (script/Build.sol, script/BuildPointers.sol or script/build.sh), so the currency check below would pass without regenerating anything."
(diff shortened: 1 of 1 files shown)


thedavidmeister:
Reviewed 9807105649055fde102e1c29955bb81b5ebbadc7: approve

Human approved for merge. All ubuntu checks pass; macos `cargo`/`check-shell` jobs pending and unrelated to a workflow YAML change; no failing checks.

Verified before merge:
- rainix itself is unaffected: no `src/generated/`, no codegen script, and no workflow calling `rainix-copy-artifacts` on this repo.
- The guard fires on 8 consumers (erc4626.words, flare, float, merkle, pyth, sol.codegen, raindex, rainlang), each fixed by `git mv script/BuildPointers.sol script/Build.sol`. rain.metadata has nothing generated and is unaffected.
- The silent-skip this replaces is demonstrated live on rainlanguage/rain.sol.codegen#31: `Regenerate pointer artifacts: skipped` with `Assert committed artifacts match freshly built: success` — green over an unregenerated tree.

PR #276, opened 2026-07-16
rainix-manual-sol-artifacts: add script + verify inputs

thedavidmeister:
Adds two optional `workflow_call` inputs to `rainix-manual-sol-artifacts.yaml`:

- **`script`** (default `script/Deploy.sol:Deploy`) — the fully qualified forge script target to run, so a consumer can run an alternate deploy script in the same repo.
- **`verify`** (default `true`) — whether to pass `--verify` to `forge script`. Disable when deploying bytecode that does not match the current source (e.g. a pinned historical release deployed from stored creation code), where Etherscan verification against current source would fail the run.

The forge invocation becomes:

```
forge script ${{ inputs.script }} -vvvvv --slow --broadcast ${{ inputs.verify && '--verify' || '' }}
```

Both inputs default to the prior behaviour, so existing callers (st0x.deploy `manual-sol-artifacts.yaml`, etc.) are unaffected.

**Motivation:** st0x.deploy is launching the audited 0.1.1 production set on Ethereum mainnet from a new `script/DeployProdV4_0_1_1.sol` that deploys the stored 0.1.1 creation bytecode. That script needs `script:` to be selected and `verify: false` because the 0.1.1 bytecode of the contracts that changed by 0.1.3 won't Etherscan-verify against the current source; those are verified manually from the 0.1.1 tag afterward.

## Summary by CodeRabbit

* **New Features**
  * Deployment workflows now support selecting a custom deployment script.
  * Deployment verification can be enabled or disabled per workflow run.
  * Deployment configuration inputs include clearer descriptions and defaults.

### DIFF
--- a/.github/workflows/rainix-manual-sol-artifacts.yaml
+++ b/.github/workflows/rainix-manual-sol-artifacts.yaml
@@ -6,8 +6,25 @@ on:
         type: string
         required: true
         description: |
-          Passed through as `DEPLOYMENT_SUITE`. `script/Deploy.sol` dispatches
-          on this to select what to deploy.
+          Passed through as `DEPLOYMENT_SUITE`. The deploy script dispatches on
+          this to select what to deploy.
+      script:
+        type: string
+        required: false
+        default: script/Deploy.sol:Deploy
+        description: |
+          Fully qualified forge script target (`path:Contract`) to run. Defaults
+          to the conventional `script/Deploy.sol:Deploy`. Override to run an
+          alternate deploy script in the same repo.
+      verify:
+        type: boolean
+        required: false
+        default: true
+        description: |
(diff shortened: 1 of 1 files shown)


thedavidmeister:
Reviewed ecea218a6c0669c5929c241bb0d1a6cab8402d39: approve

Two optional workflow_call inputs — `script` (default `script/Deploy.sol:Deploy`) and `verify` (default true) — wired into the forge command. Backward-compatible (defaults reproduce the prior invocation; verified against st0x.deploy's manual-sol-artifacts.yaml which passes neither), `${{ inputs.verify && '--verify' || '' }}` resolves correctly, YAML valid, and the unquoted `${{ inputs.script }}` stays within the existing trusted-caller boundary. CI 13/13 green, CodeRabbit clean, mergeable CLEAN.

Merge authorized by the user.

PR #280, opened 2026-07-24
feat(ci): rainix-tag-release — tag-triggered deploy-repo release

thedavidmeister:
## Why

Deploy repos and library repos need **mutually exclusive** release lifecycles, and today `st0x.deploy` (a deploy repo) is wrongly on the library one. The symptom is `testDeployTag`, red on `st0x.deploy` main for weeks — I watched it climb `0_1_8 != 0_1_9 → 0_1_26 → 0_1_27 → 0_1_28 → 0_1_29` across a single day of merges.

That test compares generated `LibProdDeployCurrent.DEPLOY_TAG` (the frozen snapshot's version) against `foundry.toml`'s `[package].version`. Under `rainix-autopublish`'s **next-version lifecycle**, `[package].version` is the *next, unpublished* version — deliberately one ahead of the registry, bumped on every content-changing merge. `DEPLOY_TAG` only advances when a deploy regenerates the pins. So the two are *designed to differ*; the test is green only in the instant after a regeneration and red on main between every merge and the next deploy.

A permanently-red check is worse than a missing one — it trains everyone to wave CI through. This PR gives deploy repos their own lifecycle where version and snapshot move only together, at release time.

## The two lifecycles

| | **Library repo** (`rainix-autopublish`) | **Deploy repo** (`rainix-tag-release`, this PR) |
|---|---|---|
| Trigger | push to `main` | push a release **tag** |
| `[package].version` | next, unpublished (ahead of registry) | last released; moves only at release |
| Per-tag snapshot | none (imports abstract surface only) | `src/generated/<tag>/` pins, frozen + append-only |
| On merge | content gate → publish → bump to next | **nothing** |

A repo is strictly one or the other.

## What a release does

Nothing on merge — a PR lands source only, main stays at the last release (its live contracts still match its pins, so the fork suite stays green). A human tags `sol-v<X>`, and this workflow runs **deploy + publish + snapshot as one act**:

**`deploy` job** — broadcasts every suite in dependency order by fanning the existing `rainix-manual-sol-artifacts` reusable over `deploy-suites` at `max-parallel: 1` (one forge run per suite — the Zoltu nonce isolation the manual dispatch already enforces). Composes the deploy reusable rather than re-implementing its broadcast/env.

**`release` job** (`needs: deploy`) — so publish/snapshot only run once the chain carries the code:

1. Sets `foundry.toml` version to `<X>` (from the tag).
2. Regenerates the **deterministic** snapshot — `address = f(bytecode)` under CREATE2, no chain access — via `snapshot-generate-cmd` (e.g. `forge script ./script/BuildPointers.sol && forge fmt`).
3. Commits it, then enforces **append-only** (`frozen-snapshots-append-only`): a release only *adds* `src/generated/<X>/`, never edits a frozen tag.
4. **Verifies the live chain matches the fresh pins** (the repo's fork suite) — confirms the deploy landed; a snapshot of addresses the chain doesn't carry fails loud *before* publishing.
5. Publishes to Soldeer at `<X>`.
6. Commits the snapshot back to `main`, so the daily drift sweep has the current release's pins.
7. Cuts the GitHub release.

`testDeployTag` then holds by construction: version and `DEPLOY_TAG` advance only in the same commit.

## Design decisions

**Sibling, not a mode on `rainix-autopublish`** — *decided*. The lifecycles share ~40% of steps (deploy-key checkout, nix setup, git config, Soldeer push, GH release) — all already in composite actions this reuses — and differ on the other ~60% (content-gate + next-version bump vs. deploy + snapshot-regen + append-only + commit-back). Guarding every autopublish step with a `publish-trigger` conditional would put two lifecycles in one file that float, raindex, and other critical packages depend on for their releases, where a subtle conditional error breaks a real publish. A sibling has zero blast radius; autopublish is untouched.

**Manual git tag is the trigger** — *decided*. The human pushes `sol-v<X>`; that fires the release. The tag ref points at the tagged source commit, while the release snapshot lands on `main` (a commit above it) and in the published Soldeer package — which is the canonical distribution consumers pin. No `workflow_dispatch`/tag-moving; the tag is exactly where the human put it.

**Deploy composes the reusable, order enforced two ways.** The `deploy` job is a matrix over `rainix-manual-sol-artifacts` (single source of truth for how a suite broadcasts). Sequencing rests on `max-parallel: 1` running the suites in listed order **and** the on-chain dep-codehash check, which fails loud if a dependent suite somehow runs before its dependency — so an ordering slip is a loud re-runnable failure (Zoltu deploys are idempotent), never silent corruption.

## Consumer change (follow-up PR on `st0x.deploy`, not here)

`package-release.yaml` flips from `on: push: branches` + `rainix-autopublish` to:

```yaml
on: { push: { tags: ['sol-v*'] } }
jobs:
  release:
    uses: rainlanguage/rainix/.github/workflows/rainix-tag-release.yaml@main
    with:
      soldeer-package: st0x-deploy
      snapshot-generate-cmd: forge script ./script/BuildPointers.sol && forge fmt
      deploy-suites: >-
        ["stox-receipt","stox-wrapped-token-vault","stox-wrapped-token-vault-beacon",
         "stox-wrapped-token-vault-beacon-set-deployer","stox-offchain-asset-receipt-vault-authorizer-v1",
         "stox-offchain-asset-receipt-vault-payment-mint-authorizer-v1","stox-corporate-actions-facet",
         "stox-receipt-vault","stox-offchain-asset-receipt-vault-beacon-set-deployer","stox-unified-deployer",
         "st0x-orchestrator","st0x-orchestrator-beacon-set-deployer"]
    secrets:
      PUBLISH_PRIVATE_KEY: ${{ secrets.PUBLISH_PRIVATE_KEY }}
      PRIVATE_KEY: ${{ secrets.PRIVATE_KEY }}
      SOLDEER_API_TOKEN: ${{ secrets.SOLDEER_API_TOKEN }}
      RPC_URL_BASE_FORK: ${{ secrets.RPC_URL_BASE_FORK }}
      RPC_URL_ETHEREUM_FORK: ${{ secrets.RPC_URL_ETHEREUM_FORK }}
      # …the fork RPCs the verify gate + deploy need, plus Etherscan keys
```

(The suite list is lifted verbatim from `st0x.deploy`'s current `manual-sol-artifacts.yaml` dispatch choices, already in dependency order.)

## Validation

- YAML parses; job graph is `deploy` (matrix, max-parallel 1, `uses: rainix-manual-sol-artifacts`) → `release` (`needs: deploy`); no duplicate keys.
- Every composite/reusable input verified against its `action.yml` / reusable `on:` (`checkout` `ssh-key`, `nix-cachix-setup`, `gh-release`, `frozen-snapshots-append-only`, `rainix-manual-sol-artifacts` `suite`/`script`/`verify`).
- rainix `pre-commit` clean (`yamlfmt`, `no-consumer-prettier`).
- Not exercised end-to-end: needs the `st0x.deploy` consumer flip and a real release tag, and the fork verify gate is currently blocked org-wide by a disabled Ankr RPC key (separate infra issue), so the consumer PR waits on that regardless.

## Summary by CodeRabbit

* **New Features**
  * Added a reusable release workflow for tag-based deployments.
  * Validates release tags and confirms they are based on the selected main branch.
  * Deploys configured suites sequentially before publishing a release.
  * Updates package versions and deployment snapshots automatically.
  * Verifies deployed networks and publishes packages to Soldeer.
  * Creates a GitHub Release and updates the main branch with release changes.

### DIFF
--- a/.github/workflows/rainix-tag-release.yaml
+++ b/.github/workflows/rainix-tag-release.yaml
@@ -1,0 +1,215 @@
+name: rainix-tag-release
+# Tag-triggered release for DEPLOY repos (deploy + publish + snapshot together),
+# the counterpart to rainix-autopublish's merge-driven publish for LIBRARY repos.
+#
+# The two lifecycles are mutually exclusive and a repo is strictly one or the
+# other:
+#
+#   * A LIBRARY repo (rainix-autopublish) runs the next-version lifecycle:
+#     [package].version is the NEXT, unpublished version, one ahead of the
+#     registry; a content change on merge publishes it and bumps to the next.
+#     Consumers import its abstract surface (interfaces/libs); it never pins a
+#     deployed address, so it carries no per-tag deploy-pin snapshot.
+#
+#   * A DEPLOY repo (this workflow) records deployed addresses. Its
+#     src/generated/<tag>/ snapshot pins the address + codehash of what it
+#     deployed, frozen so consumers can rely on them (enforced by the
+#     frozen-snapshots-append-only gate). [package].version is the LAST released
+#     version, and moves ONLY at release time, in lockstep with the snapshot it
+#     describes.
+#
+# Running the next-version lifecycle on a deploy repo is the bug this exists to
+# remove: autopublish bumps [package].version on every merge, while the frozen
+# DEPLOY_TAG only advances at deploy time, so a version-vs-DEPLOY_TAG identity
+# test (e.g. `testDeployTag`) is red on main between every merge and the next
+# deploy — trained to be ignored, which is how a real regression rides through.
+#
+# Here nothing moves on merge: a PR lands source only, main stays at the last
+# release (its live contracts still match its pins), and a human TAG is the sole
+# release trigger. The tag names the version; this workflow regenerates the
+# snapshot for it, verifies the live chain against the fresh pins, publishes to
+# Soldeer, and commits the new (append-only) snapshot back to main so the daily
+# drift sweep always has the current release's pins to check.
+#
+# The on-chain DEPLOY itself stays the repo's existing per-network, human-driven
+# rainix-manual-sol-artifacts dispatch, run BEFORE tagging: the tag's verify gate
+# passes precisely because the deploy already happened. Folding that dispatch
+# into this workflow (so one tag deploys every suite in dependency order, then
+# publishes) is the intended next step — see the PR description.
+on:
+  workflow_call:
+    inputs:
+      soldeer-package:
+        description: Soldeer registry package name to publish (e.g. st0x-deploy).
+        required: true
+        type: string
+      tag-prefix:
+        description: >-
+          Prefix stripped from the pushed tag to derive the release version, e.g. `sol-v` turns tag `sol-v0.1.29` into version `0.1.29`. The caller restricts which tags trigger the release via its own `on: push: tags` filter; this only parses the version out of the ref.
+        required: false
+        type: string
+        default: sol-v
+      snapshot-generate-cmd:
+        description: >-
+          Command that regenerates the deploy-pin snapshot from the (deterministic) bytecode into src/generated/<tag>/, DEPLOY_TAG and any pointer libs, then formats. Run after [package].version is set to the release version, so the generated tag matches it. e.g. `forge script ./script/BuildPointers.sol && forge fmt`.
+        required: true
+        type: string
+      test-cmd:
+        description: >-
+          Pre-publish verification gate. Run against the regenerated snapshot; for a deploy repo this is the fork suite that reads the live chain and asserts it matches the fresh pins, so a release that snapshots addresses the chain does not actually carry fails loud BEFORE publishing. Default `forge test`.
+        required: false
+        type: string
+        default: forge test
+      main-branch:
+        description: The branch the release snapshot is committed back to.
+        required: false
+        type: string
+        default: main
+    secrets:
+      PUBLISH_PRIVATE_KEY:
+        # A deploy key whose push events (unlike GITHUB_TOKEN pushes) trigger the
+        # downstream git-clean / rainix-sol workflows on the commit-back to main.
+        required: false
+      CI_GIT_EMAIL:
+        required: false
+      CI_GIT_USER:
+        required: false
+      SOLDEER_API_TOKEN:
+        required: false
+      CACHIX_AUTH_TOKEN:
+        required: false
+      RPC_URL_ARBITRUM_FORK:
+        required: false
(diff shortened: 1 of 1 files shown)


thedavidmeister:
Addressed the CodeRabbit findings in f1fa5a9:

- **🔴 Tag provenance (was L188):** new `guard` job requires the tagged commit to be an ancestor of the release branch (`git merge-base --is-ancestor "$GITHUB_SHA" origin/$MAIN`) before `deploy` or `release` runs — a tag cut from an unmerged branch is now rejected, so the rebase can't replay unreviewed commits onto main and the deploy can't broadcast unreviewed bytecode. All refs come from built-in env vars, not `github.ref` interpolation (**🔒 was L190**).
- **🔴 Worktree restore (was L237):** removed `git checkout -- .` from the commit step — it reverted the `foundry.toml` version bump and any regenerated tracked file, which would have broken the version/snapshot lockstep. The one devShell leftover is already hidden via `.git/info/exclude`.
- **🟠 VERSION validation (was L223):** the parsed version must match `MAJOR.MINOR.PATCH`; this rejects junk tags (`sol-vfoo`) and, since semver carries no `&`/`\`/`/`, removes any sed-injection into `foundry.toml`. Also fails loud if no `[package] version` line was updated.
- **🟠 SOLDEER_API_TOKEN (was L118):** `required: true` — this workflow always publishes, so a missing token should fail at setup, not after the deploy.
- **🔒 id-token (was L172):** dropped the unused `id-token: write`; only `contents: write` remains (Soldeer uses `SOLDEER_API_TOKEN`, the release uses `GITHUB_TOKEN`).

**Not fixed — documented follow-up:** publish-then-commit-back resumability (was L281). If Soldeer publish succeeds but the push to main fails, the package is published without its snapshot on main. This is the atomicity limitation already noted in the PR description; a proper fix needs durable release state and is a heavy lift for a follow-up, not this PR.

@coderabbitai review

thedavidmeister:
Reviewed f1fa5a97f4c591b359f7ebea68d6f794eadded1f: approve

All 12 rainix checks green on this sha. The two CodeRabbit Criticals (tag provenance, worktree restore) and three Majors (VERSION validation, SOLDEER_API_TOKEN required, unused id-token) are fixed in f1fa5a9 — those threads are now outdated. The one remaining finding, publish/commit-back resumability, is a documented heavy-lift follow-up (a rare partial-failure recovery path), not a defect in what lands here; thread resolved with that note. Reviewed my own authored change adversarially: the guard job's ancestry check closes the unmerged-tag hole, and removing `git checkout -- .` restores the version/snapshot lockstep this PR exists to guarantee.

PR #282, opened 2026-07-24
tag-release: drop the deploy job; deploys stay manual

thedavidmeister:
Follow-up to #280. Decouples the on-chain deploy from the tag publish.

## Why

A deploy is flaky by nature — per-network, funds-dependent, and at the mercy of RPC archive/fork behaviour. A retry-prone operation like that must not gate a **one-shot** tag publish: a single transient RPC failure blocks the entire release.

This isn't hypothetical — it's what happened on the first real run. Cutting `sol-v0.1.6` on `rain.factory.deploy` failed in the folded-in deploy job:

```
VM::createSelectFork("arbitrum") → HTTP 400
{"error":{"message":"Unknown state. First available state is 1","code":27}}
```

The deploy forked Arbitrum against a load-balanced RPC and got routed to a backend without the state — and took the whole release down with it, even though nothing was wrong with the code or the publish. (The fork *tests* passed on the same endpoint minutes earlier; load-balanced routing is nondeterministic.)

## The change

The deploy goes back to being the repo's own human-driven `rainix-manual-sol-artifacts` dispatch, run **before** tagging. This workflow no longer broadcasts — it **attests**:

- `guard` → `release` (the `deploy` job is gone; `release` now `needs: guard`).
- The `verify` step (the repo's fork suite) checks **prod exists at the freshly regenerated pins**. That *is* the confirmation the manual deploy landed. If the pins don't resolve on-chain, verify fails and nothing publishes.
- A transient fork/RPC failure in verify just fails the release — retry the tag. It's a read-only check, not a broadcast, so it ships nothing and can't leave a partial deploy.

Removes the `deploy` job, the `deploy-suites`/`deploy-script`/`deploy-verify` inputs, and the deploy-only secrets (`PRIVATE_KEY`, `EXPLORER_VERIFICATION_KEY`, `CI_DEPLOY_*` etherscan keys). The fork-RPC secrets stay — verify still forks.

## Consumer follow-up

`rain.factory.deploy`'s `package-release.yaml` passes `deploy-suites`, which this removes — a companion one-line edit drops that input. (Separate push; that repo's next release tag uses the manual-deploy-then-tag flow.)

## Validation

- YAML valid; jobs are `guard` → `release` (`needs: guard`); no `deploy` job, no deploy inputs/secrets; fork-RPC secrets retained.
- rainix `pre-commit` clean (yamlfmt).

[...]

Issue #283, opened 2026-07-24
rainix-sol-artifacts CI red since 2026-07-24: ETH_RPC_URL resolves empty (org secret/var, not a code regression)


thedavidmeister:
## Symptom

`Rainix CI` → `rainix-sol-artifacts` has failed on every push to `main` since the merge of #279 (commit `1086e28`, 2026-07-24T11:59:59Z), and continues to fail on every commit since (`06fa7b6`, `8583c77`, ...). Each run retries the deploy 10 times over ~40s before giving up with:

```
Error: Internal transport error: Connection refused (os error 111) with /home/runner/work/rainix/rainix/test/fixture/
```

That path looks like a missing local fixture/mock-RPC service, but it isn't one — see root cause below.

## Root cause

`.github/workflows/test.yml` runs the job with:

```yaml
env:
  ETH_RPC_URL: ${{ secrets.CI_DEPLOY_SEPOLIA_RPC_URL || vars.CI_DEPLOY_SEPOLIA_RPC_URL }}
```

and `rainix-sol-artifacts` (in `flake.nix`) invokes `forge script ... --rpc-url "${ETH_RPC_URL}"`.

`ETH_RPC_URL` is resolving to an **empty string**. Confirmed directly from the job's own env dump, printed before any retry logic runs, identically in all three failing runs:

- job `89475572223` (run 30091492683, commit `1086e28`): `ETH_RPC_URL: ` (blank)
- job `89482744436` (run 30093743505, commit `06fa7b6`): `ETH_RPC_URL: ` (blank)
- job `89492033946` (run 30096584996, commit `8583c77`): `ETH_RPC_URL: ` (blank)

With `--rpc-url ''`, forge can't parse it as a URL and falls back to treating it as a local path — the job's cwd (`test/fixture/`, set via `defaults.run.working-directory`) — then fails to connect to that "endpoint." Hence the misleading "connection refused ... test/fixture/" text: there is no fixture server involved, and nothing under `test/fixture/` changed or broke.

## Why this isn't a code regression

- `.github/workflows/test.yml` is byte-identical between the last known-good run and every failing run since (`git diff 409759f..8583c77 -- .github/workflows/test.yml` is empty).
- The last successful `rainix-sol-artifacts` run was 2026-07-16 (run `29506458832`, job `87648364549`): the job log shows `--rpc-url ***` (masked, i.e. **non-empty**) and `SIMULATION COMPLETE`.
[...]

PR #285, opened 2026-07-24
fix(tag-release): map fork RPC secrets to the env names foundry reads

thedavidmeister:
Fixes a real bug in `rainix-tag-release`'s verify step (introduced in #280, unchanged by #282), caught when `rain.factory.deploy`'s first release run failed.

## Bug

The verify step named its env vars after the **secrets** — `RPC_URL_ARBITRUM_FORK`, `RPC_URL_BASE_FORK`, … — but `foundry.toml`'s `[rpc_endpoints]` read `${ARBITRUM_RPC_URL}`, `${BASE_RPC_URL}`, …. So the vars foundry actually needs were never set, and every fork in the verify suite died with:

```
vm.createSelectFork: environment variable `ARBITRUM_RPC_URL` not found
```

The test reusable (`rainix-sol-test.yaml`) maps them correctly; the tag-release verify step didn't. Until this lands, no tag-release can pass verify → nothing publishes.

## Fix

Map each `RPC_URL_<NETWORK>_FORK` secret to the `<NETWORK>_RPC_URL` env name foundry expects (mirroring `rainix-sol-test.yaml`), with the `|| vars.*` fallback the rest of the org's CI uses. yamlfmt clean.

### DIFF
--- a/.github/workflows/rainix-tag-release.yaml
+++ b/.github/workflows/rainix-tag-release.yaml
@@ -227,13 +227,18 @@ jobs:
         # pins, so a snapshot of addresses the chain does not carry never gets
         # published. A transient fork/RPC failure here just fails the release
         # (retry the tag); it publishes nothing and is not a broadcast.
+        # foundry.toml's [rpc_endpoints] read ${<NETWORK>_RPC_URL}; the secrets are
+        # named RPC_URL_<NETWORK>_FORK. Map each to the env name foundry expects
+        # (mirrors rainix-sol-test.yaml), with the vars.* fallback the rest of the
+        # org's CI uses. Naming them after the secret (as before) left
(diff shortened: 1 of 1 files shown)


thedavidmeister:
Reviewed 97fbfc7c0040512925587f67c06c466eef1a98bf: approve

CodeRabbit rate-limited (not reviewing). Own review: the change maps each RPC_URL_<NETWORK>_FORK secret to the <NETWORK>_RPC_URL env name foundry.toml reads, mirroring rainix-sol-test.yaml, with the vars.* fallback — trivially correct, and the exact fix for the createSelectFork 'env var not found' that failed rain.factory.deploy's verify.

Green-by-rule: the only red is rainix-sol-artifacts, inherited from main (the eth-sepolia RPC-secret issue in #283/#284), unrelated to this env-name change.

PR #289, opened 2026-07-25
Fail over fork RPC endpoints with an archive-aware preflight

### LINKED ISSUE #288: Fork RPC has no failover: one vendor's exhausted plan quota reds CI org-wide
## The problem

The org secret `RPC_URL_ARBITRUM_FORK` points at drpc (`lb.drpc.live`), whose plan quota is exhausted:

```
error code -32001: You've reached the usage limit for your current plan
```

One commercial failure at one vendor reddened CI across the org: `rain.extrospection` and `rain.factory.deploy` on the quota error, `cyclo.sol` on a related archive error.

**Foundry has no failover.** `[rpc_endpoints]` maps one alias to exactly one URL, and `--fork-retries` / `--rpc-timeout` only retry the *same* endpoint, which cannot help against a deterministic error like an exhausted plan quota. (Verified on forge 1.7.2-nightly.) `flake.nix`'s `rainix-sol-artifacts` has the same shape — an `until do_deploy; do sleep 5` loop that re-hits `${ETH_RPC_URL}` ten times before failing anyway.

The reusables map secrets to env with `secrets.X || vars.X`, which is a fallback for a **missing** value, not a **failing** endpoint. Nothing in the org can currently route around a dead upstream.

## The archive trap

Different repos need different things from the *same* shared secret, and the requirement is invisible at the call site:

- `rain.factory.deploy` does `vm.createSelectFork(LibRainDeploy.ARBITRUM_ONE)` with no block pin — latest state only, a pruning node is fine.
- `cyclo.sol` forks Arbitrum at a pinned historical block and fails with `It looks like you're trying to fork from an older block with a non-archive node which is not supported.`
- Worse, `LibRainDeploy.isStartBlock` forks at **latest** and then `vm.rollFork()`s backwards, so grepping for the two-arg `createSelectFork(url, block)` form misses the archive requirement on base_sepolia and polygon entirely.

Deepest live pin per network across the org today:

| network | deepest requirement | source |
| --- | --- | --- |
| arbitrum | 280,000,000 (~20 months) | `rain.tofu.erc20-decimals` |
| base | block 0 (genesis binary search) | `rain.deploy` `findDeployBlock` |
| base_sepolia | 38,683,088 (~4.5 months) | `rain.metadata` metaboard start block |
| flare | 31,843,105 (~23 months) | `rain.flare`, ~19 test files |
| polygon | 82,855,948 (~6.4 months) | `rain.metadata` metaboard start block |
| ethereum | none — every fork is at latest | — |

So **a health check that only calls `eth_blockNumber` will happily select a pruning node that passes the check and then fails the suite.** Measured at cyclo's pin: `arb1.arbitrum.io/rpc` NOT archive, `arbitrum-one-rpc.publicnode.com` NOT archive (archive gated behind a personal token), `arb-pokt.nodies.app` OK.

Two further measured traps:

- `eth_getCode` is **also** insufficient. `base.meowrpc.com` serves historical code fine but answers *every* `eth_call` with `-32000 "The method eth_call is not supported."` A code-only probe selects it and the suite dies later.
- **A single probe sample is not enough.** `1rpc.io` and `drpc.org` load-balance over heterogeneous backend pools: `1rpc.io/arb` passed the historical probe 1 time in 5 (the other four were `missing trie node`), and `1rpc.io/base` was archive on one sample and `state is pruned` twenty minutes later. A one-shot preflight qualifies such an endpoint and the suite still dies halfway through, because forge issues thousands of historical reads.

## Design

A composite action under `.github/actions/` (house pattern: `nix run "path:…#rainix-static"`, so the check version always matches the action version), delegating to a new `rainix-static rpc-preflight` subcommand per the "tooling is Rust, never bash logic" rule. Wired into the three reusables that set fork RPC env, replacing the static `secrets.X || vars.X` mapping. rainix owns shared CI, so consumers inherit the fix without touching their own workflows.

### Merged candidate pool, not a precedence chain

`RPC_URL_<NETWORK>_FORK` as a **secret** and as a **variable** each hold a **newline-separated list** of URLs. The candidate pool is the **concatenation** of both lists plus hardcoded public archive defaults. A URL in the variable is a real candidate even when the secret is also set — a non-empty earlier source never causes a later one to be skipped, it only orders them:

| order | source | holds |
| --- | --- | --- |
| 1 | secret `RPC_URL_<NET>_FORK` | keyed/paid URLs (masked) |
| 2 | variable `RPC_URL_<NET>_FORK` | public keyless URLs (visible, greppable) |
| 3 | hardcoded defaults | measured keyless archive endpoints |

That ordering prefers the paid/dedicated endpoint, falls back to the org's curated public list, and lands on the hardcoded safety net only when both are exhausted. Keeping keyed URLs in the secret and keyless ones in the variable is the whole point of merging them: a public archive endpoint can back up a keyed one **without putting a non-secret into a secret**, where masking makes logs unreadable for no security benefit.

Newline is the separator because it cannot occur inside a URL (no escaping), GitHub's secret/variable editors take multi-line values natively, and the runner registers each *line* of a multi-line secret as a separately masked value. A single bare URL is a one-element list — exactly what these secrets hold today, so this is backwards compatible.

### Archive-aware probe

Per candidate, fail-cheapest-first:

1. `eth_chainId` — a candidate for the wrong chain would silently fork the wrong network, which is worse than a red job.
2. `eth_getBalance` at the probe block — historical **account** state, which is what forge's fork backend fetches first and what a pruning node refuses. (The `arb1` failure is on the sequencer account, not the token.)
3. `eth_call` at the probe block — historical **execution**, necessary on top of (2) because of the meowrpc class of host.

Repeated 3 times consecutively, all must pass, because of the heterogeneous-pool finding above. The probe block is the **deepest block any repo in the org pins for that network** — the question is not "is this an archive node" but "can it serve *my* fork block", and probing shallower than the deepest live pin qualifies a node that then dies mid-suite. Ethereum and HyperEVM are latest-only in every consumer so they are not held to the archive bar; neither are deploy/broadcast paths, which only ever read head state (`RPC_URL_<NET>_FORK` is a misnomer there — the same secret feeds a real `forge script --broadcast`).

### No candidate URL is ever printed

Logs name the **source** and a **typed reason**, never a URL:

```
rpc-preflight: arbitrum: secret[0] rejected: quota exhausted / rate limited (rpc error -32001)
rpc-preflight: arbitrum: SELECTED variable[1] (chain 42161, archive at block 280000000, 3/3 samples)
```

This has to be structural rather than a redaction pass, because a keyed URL that arrives via `vars.` (a misconfiguration that will happen) is not auto-masked by the runner.

### Scope containment

Only networks the repo actually references are probed, and a network with no candidates at all is left exactly as it is today — so a repo that sets no secret, and a repo that uses one network, are both unaffected in every way except getting a working endpoint.

## Open question

"All used" is implemented as *all are candidates the preflight will actually try*, because Foundry consumes exactly one URL per alias. True simultaneous multi-upstream load balancing would need a proxy in front of forge, not a URL list — flagging in case that was the intent.

thedavidmeister:
Closes #288.

Foundry maps one `[rpc_endpoints]` alias to exactly one URL and `--fork-retries` only retries that same URL, so a deterministic upstream failure — the exhausted drpc plan quota that reddened `rain.extrospection` / `rain.factory.deploy`, a pruning node, a dead host — cannot be recovered inside `forge`. This recovers it one step earlier: a preflight picks a candidate that is healthy *right now* and binds it to `<NETWORK>_RPC_URL` before forge starts.

## What's here

- **`rainix-static rpc-preflight`** (`rainix-static/src/rpc_preflight.rs`) — all the logic, per the "tooling is Rust, never bash" rule. 20 unit tests, run inside the nix build.
- **`.github/actions/rpc-preflight/action.yml`** — house pattern: `nix run "path:$GITHUB_ACTION_PATH/../../..#rainix-static"`, so the check version always matches the action version regardless of any `RAINIX_SHA` the caller pins.
- Wired into the three reusables that set fork RPC env — `rainix-sol-test.yaml`, `rainix-tag-release.yaml`, `rainix-manual-sol-artifacts.yaml` — **replacing** the static `secrets.X || vars.X` mapping in each.
- `flake.nix`: `rainix-static` is now wrapped with the pinned `curl` and `git` plus a CA bundle. The composite runs `nix run` *outside* any devshell, so ambient PATH is whatever the runner image ships. Verified hermetic — the nix-built binary completes a full selection under `env -i PATH=/nonexistent`.

## Merged pool, not a precedence chain

Per the ruling — *"i didn't say fallback if not set i said merge into a list of urls that are all used"*. The secret and the variable each hold a **newline-separated list**, and the candidate pool is the **concatenation** of both plus hardcoded defaults:

| order | source | holds |
| --- | --- | --- |
| 1 | secret `RPC_URL_<NET>_FORK` | keyed/paid URLs (masked) |
| 2 | variable `RPC_URL_<NET>_FORK` | public keyless URLs (visible, greppable) |
| 3 | hardcoded defaults | measured keyless archive endpoints |

A URL in the variable is tried **even when the secret is set** — a non-empty earlier source never causes a later one to be skipped, it only orders them. Ordering is secret-first because that is the paid/dedicated endpoint; public ones are the safety net. In the demo below `secret[0]`, `secret[1]` and `variable[0]` are all rejected and `variable[1]` is selected, which is only reachable under the merged reading.

**Separator = newline**, because it cannot occur inside a URL (no escaping), GitHub's secret/variable editors take multi-line values natively, and the runner registers each *line* of a multi-line secret as a separately masked value. A single bare URL parses as a one-element list, which is exactly what these secrets hold today — **no configuration change is required for this PR to be an improvement**; the hardcoded defaults alone already fix the outage. `#` starts a comment so a dead candidate can be parked with a note.

`Foundry consumes exactly one URL per alias`, so "all used" is implemented as *all are candidates the preflight will actually try*. See the open question at the bottom.

## The probe is archive-aware, and that is the load-bearing part

`eth_blockNumber` is useless here: it passes on a pruning node that then fails the suite. So is `eth_getCode` alone — `base.meowrpc.com` serves historical code but answers *every* `eth_call` with "method not supported". Per candidate, fail-cheapest-first:

1. `eth_chainId` against the expected chain id — a wrong-chain candidate would *silently fork the wrong network*, worse than a red job.
2. `eth_getBalance` at the probe block — historical **account** state, what forge's fork backend fetches first.
3. `eth_call` at the probe block — historical **execution**, and the result must be a full 32-byte word (a `0x` answer means the node served empty state).

**3 consecutive full passes required.** `1rpc.io/arb` passed the historical probe 1 time in 5 and `1rpc.io/base` flipped from archive to `state is pruned` within twenty minutes — these hosts round-robin over a mix of archive and pruning backends, so a single sample qualifies an endpoint that dies partway through a suite issuing thousands of historical reads.

**Probe block = the deepest block any repo in the org pins for that network**, because the question is not "is this an archive node" but "can it serve *my* fork block":

| network | probe block | why |
| --- | --- | --- |
| arbitrum | 280,000,000 | `rain.tofu.erc20-decimals` (~20 months) |
| base | 1 **and** 39,000,000 | `rain.deploy` `findDeployBlock` binary-searches from genesis; the second block stops a node holding only early state from passing |
| base_sepolia | 38,000,000 | `rain.metadata` metaboard start block |
| flare | 31,000,000 | `rain.flare`, ~19 test files (~23 months) |
| polygon | 82,000,000 | `rain.metadata` metaboard start block |
| ethereum | latest | every Ethereum fork in the org is at latest; a pruning node is fine and should not be excluded |
| hyperevm | latest | latest-only in every consumer |

Note base_sepolia and polygon have **no** two-arg `createSelectFork(url, block)` anywhere — their archive requirement is hidden inside `LibRainDeploy.isStartBlock`, which forks at latest and then `vm.rollFork()`s backwards.

`archive: 'false'` on the deploy path (`rainix-manual-sol-artifacts`): a broadcast only reads head state, so requiring archive there would reject a healthy pruning endpoint for nothing. The `_FORK` suffix on the secret is a misnomer on that path.

## How the no-leak guarantee works

Not a scrubbing pass — those get bypassed by the next edit. It is structural, in three layers:

1. **`struct Url(String)` has no `Display`, and its `Debug` prints `<redacted>`.** No format string anywhere in the crate can put a URL on stdout or stderr, including a future edit that adds a `{:?}` to a log line. The inner value is reachable only via `expose()`, which has exactly **three** call sites: curl's argv, the write to `$GITHUB_ENV`, and the `::add-mask::` command in (3) — whose entire purpose is that the runner redacts the value it is handed, including in the command line itself. There is deliberately **no stdout mode**: invoked without a file to write to, the subcommand fails rather than falling back to printing.
2. **Every log line is a `Source` plus a `Reason`, both closed enums over fixed text.** `Reason`'s `Display` is fixed strings plus integers we produced ourselves (a JSON-RPC code, an HTTP status, a curl exit, a chain id, a block). Upstream response bodies are **classified and then dropped** — an endpoint that echoes the request URL back inside its error message, or serves an HTML error page containing it, cannot leak it. curl's stderr is `Stdio::null()`ed because curl writes the request host into its own error text.
3. **`::add-mask::` on any secret-sourced selection** before it reaches `$GITHUB_ENV`. Redundant with the runner's own per-line masking of multi-line secrets, but free — and it does not help for the case that actually matters, which is a keyed URL misconfigured into `vars.` (not auto-masked). That case is covered by (1) and (2).

Candidate URLs also travel to the composite as **env vars, not `with:` inputs**, because GitHub renders a composite's resolved inputs into the log.

There is a test asserting `Url`'s debug is `<redacted>` and one asserting no `Reason` variant renders a `http://` or `https://`.

## Blast radius containment

- **Only networks the repo references are probed.** A demand scan over `git ls-files` looks for `<NET>_RPC_URL` (the foundry.toml alias form) *and* `RPC_URL_<NET>_FORK` (the raw-secret form `cyclo.sol` reads via `vm.envString`). An unmentioned network is never probed, never exported, and can never fail the job. A test asserts `BASE_SEPOLIA_RPC_URL` does not drag `base` in.
- **A network with no candidates at all is a skip, not a failure** — byte-identical to today's behaviour (env stays unset).
- The selected URL is exported under **both** `<NET>_RPC_URL` and `RPC_URL_<NET>_FORK`. The second matters: once the secret can hold a list, repos reading the raw secret name would otherwise get a multi-line string `createSelectFork` cannot use.

## Riding along

`RPC_URL_HYPEREVM_FORK` was declared in `rainix-manual-sol-artifacts.yaml` only, so a repo with `hyperevm` in `[rpc_endpoints]` could deploy but its fork tests and tag-release verification both failed "env var not found". Added to `rainix-sol-test.yaml`, `rainix-sol.yaml` (declaration **and** the explicit forward — that forward is a hard allowlist, `secrets: inherit` upstream does not get past it) and `rainix-tag-release.yaml`. README's documented env list was also missing `ETHEREUM_RPC_URL` and `HYPEREVM_RPC_URL`.

## Demonstration

Local end-to-end run against a synthetic consumer repo whose `foundry.toml` declares `arbitrum` and whose test reads `RPC_URL_BASE_SEPOLIA_FORK` directly. The secret list is `[an exhausted-plan upstream with ?dkey=SUPERSECRETKEY123, a pruning node]`, the variable list is `[publicnode (archive token-gated), thirdweb]`. Polygon is configured but **not referenced by the repo**. The first two upstreams are local mocks replaying the verbatim drpc and geth error bodies; everything else is a real public endpoint.

```
rpc-preflight: arbitrum: secret[0] rejected: quota exhausted / rate limited (rpc error -32001)
rpc-preflight: arbitrum: secret[1] rejected: not archive: no historical state at block 280000000 (rpc error -32000)
rpc-preflight: arbitrum: variable[0] rejected: not archive: no historical state at block 280000000 (rpc error -32602)
::warning::rpc-preflight: arbitrum fell back to variable[1] after 3 unhealthy candidate(s)
rpc-preflight: arbitrum: SELECTED variable[1] (chain 42161, archive at block 280000000, 3/3 samples)
rpc-preflight: base_sepolia: SELECTED default[0] (chain 84532, archive at block 38000000, 3/3 samples)
rpc-preflight: clean
exit=0

--- $GITHUB_ENV written by the step ---
ARBITRUM_RPC_URL=https://42161.rpc.thirdweb.com
RPC_URL_ARBITRUM_FORK=https://42161.rpc.thirdweb.com
BASE_SEPOLIA_RPC_URL=https://sepolia.base.org
RPC_URL_BASE_SEPOLIA_FORK=https://sepolia.base.org

=== LEAK CHECK: does anything the step printed contain the secret's key? ===
clean: 'SUPERSECRETKEY123' appears 0 times in the step log
host substring occurrences in log: 0
```

Every rejection is named by source with a typed reason. Polygon is absent because the repo does not reference it. base_sepolia had nothing configured and landed on a hardcoded default — the case of a repo that sets no secret at all today.

**Total-failure path** — every candidate for a required network unreachable (an outbound proxy pointed at a closed port, so the hardcoded defaults die too):

```
rpc-preflight: arbitrum: secret[0] rejected: unreachable (curl exit 7)
rpc-preflight: arbitrum: variable[0] rejected: unreachable (curl exit 7)
rpc-preflight: arbitrum: default[0] rejected: unreachable (curl exit 7)
rpc-preflight: arbitrum: default[1] rejected: unreachable (curl exit 7)
rpc-preflight: arbitrum: default[2] rejected: unreachable (curl exit 7)
::error::rpc-preflight: arbitrum: no healthy endpoint among 5 candidate(s) for chain 42161 (archive at block 280000000). Every candidate and why it was rejected is listed above by SOURCE; add a working URL to the RPC_URL_ARBITRUM_FORK secret (keyed) or variable (public, newline-separated).
rpc-preflight: base_sepolia: default[0] rejected: unreachable (curl exit 7)
...
::error::rpc-preflight: no healthy RPC endpoint for: arbitrum, base_sepolia
exit=1
```

Note it does **not** stop at the first dead network — every required network is reported, so one run tells you everything that needs fixing. The secret here was `https://keyed.example/?dkey=SUPERSECRETKEY123`; neither the key, nor the host, nor the variable's host appears anywhere in that output.

**Closing the loop with real forge** — same pin `rain.extrospection` uses, isolated `HOME` so there is no warm fork cache:

```
### FORGE, with the URL the preflight REJECTED as not-archive (public arb1) ###
Ran 1 test for test/Fork.t.sol:ForkTest
[FAIL: EVM error; database error: failed to get account for 0xA4b000000000000000000073657175656e636572: server returned an error response: error code -32000: metadata is not found, 402243830] testForkAtPinnedBlock() (gas: 0)
Suite result: FAILED. 0 passed; 1 failed; 0 skipped

### FORGE, with the URL the preflight SELECTED ###
Ran 1 test for test/Fork.t.sol:ForkTest
[PASS] testForkAtPinnedBlock() (gas: 16218)
Suite result: ok. 1 passed; 0 failed; 0 skipped
```

That `failed to get account for 0xA4b0…sequencer` is precisely the cryptic mid-suite error this replaces with `secret[0] rejected: quota exhausted`.

## Checks

- `cargo test` — 34 pass (20 new), `cargo fmt --check` and `cargo clippy -D clippy::all` clean.
- `nix develop -c pre-commit run --all-files` — yamlfmt, prettier-rainix, shellcheck, statix, taplo, nil, nixfmt, deadnix, denofmt all pass. The `rustfmt-conditional` hook fails **identically on pristine `main`** (`cargo metadata` looks for a `Cargo.toml` at the repo root, which rainix does not have); it is not exercised by CI, where `cargo-fmt` is off PATH and the hook no-ops. Not touched here — separate issue.

## Open question

`Foundry consumes exactly one URL per alias`, so "all used" is implemented as *every URL in either list is a candidate the preflight will actually try*, which is what the words say. True simultaneous multi-upstream load balancing would need a proxy sitting in front of forge, not a URL list. Flagging rather than silently choosing, in case that was the intent.

Two follow-ups deliberately **not** in this PR, one of which is **red on `main` right now**:

- **sepolia is not covered, and it is the reason `main` is red.** `rainix (ubuntu-latest, rainix-sol-artifacts)` has failed on every `main` run since 2026-07-24 (green on 07-16), including `aa789b99`, which this branch is cut from. The log is the same failure class this PR fixes, in its purest form:

  ```
  ++ forge script script/Deploy.sol:Deploy -vvvvv --slow --rpc-url '' --etherscan-api-key *** --resume
  Error: Internal transport error: Connection refused (os error 111)
  Deploy failed, retrying in 5 seconds... (attempt 7)
  ...
  Deploy failed after 10 attempts, aborting.
  ```

  `--rpc-url ''` — `secrets.CI_DEPLOY_SEPOLIA_RPC_URL || vars.CI_DEPLOY_SEPOLIA_RPC_URL` resolves empty, and `flake.nix`'s `until do_deploy; do sleep 5` loop then re-hits the empty URL ten times over ~50s before failing anyway. A candidate pool with a hardcoded default would make this green with **zero** configuration.

  I did not fold sepolia in, because doing so makes two decisions that are yours, not mine: (1) sepolia is the one network outside the `RPC_URL_<NET>_FORK` scheme — secret `CI_DEPLOY_SEPOLIA_RPC_URL` → env `ETH_RPC_URL` *and* `CI_DEPLOY_SEPOLIA_RPC_URL`, with `flake.nix` hardcoding `--rpc-url ${ETH_RPC_URL}`, so folding it in either enshrines that naming or migrates it; and (2) that job **broadcasts** with `DEPLOYMENT_KEY`, so having a hardcoded default pick the endpoint for a real transaction is a materially bigger call than picking one for a fork read. Say which way you want it and it is a small follow-up. It is also block-pinned by a repo variable (`CI_FORK_SEPOLIA_BLOCK_NUMBER`), so its archive depth is configuration I cannot see from here.

- **`flake.nix`'s `rainix-sol-artifacts` retry loop** is the other half of that log: ten retries with `--resume` against the same endpoint is structurally identical to Foundry's `--fork-retries` and just as useless against a deterministic failure. Classifying the error once and either failing fast or moving to another candidate is the same fix as this PR, applied to the deploy task.

### DIFF
--- a/.github/actions/rpc-preflight/action.yml
+++ b/.github/actions/rpc-preflight/action.yml
@@ -1,0 +1,27 @@
+name: rpc-preflight
+description: >-
+  Selects a working fork RPC endpoint per network and exports it as <NETWORK>_RPC_URL for the steps that follow, so a dead upstream fails over instead of reddening every fork suite in the org. Foundry maps one [rpc_endpoints] alias to exactly one URL and its --fork-retries only retries that same URL, so a deterministic upstream failure (plan quota exhausted, node is not archive, host is gone) cannot be recovered inside forge — it has to be recovered before forge starts. Candidates are the MERGED pool of the RPC_URL_<NETWORK>_FORK secret (newline-separated list of keyed/paid URLs), the RPC_URL_<NETWORK>_FORK variable (newline-separated list of public keyless URLs) and hardcoded public archive defaults; every entry in every source is a real candidate, the sources only set the order. Health is an archive-aware probe (chain id, then historical account state and a historical eth_call at the deepest block any repo in the org pins for that network) repeated until it passes consecutively, because a latest-only check happily selects a pruning node that then dies mid-suite. Only networks the repo actually references are probed; a network with no candidates at all is left exactly as it is today.
+inputs:
+  archive:
+    description: >-
+      Whether the endpoint must serve historical state. 'true' for fork test suites (they roll back months of blocks). 'false' for deploy/broadcast paths, which only ever read head state — requiring archive there would reject a healthy pruning endpoint for no benefit.
+    required: false
+    default: 'true'
+runs:
+  using: composite
+  steps:
+    - name: Select a healthy fork RPC per network
+      shell: bash
+      # Candidate URLs are NOT passed as `with:` inputs on purpose: GitHub
+      # renders a composite's resolved inputs into the log, which would print a
+      # keyed URL verbatim. They arrive as env vars set by the calling reusable
+      # (RAINIX_RPC_SECRET_<NET> / RAINIX_RPC_VARS_<NET>) and are read straight
+      # out of the environment by the binary, which never prints one.
+      run: |
+        set -euo pipefail
+        # Single source of truth: the Rust rainix-static binary (its unit tests
+        # run inside the nix build). The path: flake ref runs it from this
+        # composite's own checkout, so the check version always matches the
+        # action version regardless of any RAINIX_SHA the caller pins.
+        nix run "path:$(cd "$GITHUB_ACTION_PATH/../../.." && pwd)#rainix-static" -- \
+          rpc-preflight ${{ inputs.archive == 'true' && '' || '--no-archive' }}
--- a/.github/workflows/rainix-manual-sol-artifacts.yaml
+++ b/.github/workflows/rainix-manual-sol-artifacts.yaml
@@ -107,18 +107,38 @@ jobs:
         if: hashFiles('soldeer.lock') != ''
         run: nix develop github:rainlanguage/rainix/${{ env.RAINIX_SHA }}#sol-shell -c forge soldeer install
       - run: nix develop github:rainlanguage/rainix/${{ env.RAINIX_SHA }}#sol-shell -c forge selectors up --all
+      # archive: false — this is a broadcast, so it only ever reads head state.
+      # Requiring historical state here would reject a perfectly good pruning
+      # endpoint; the RPC_URL_<NET>_FORK name is a misnomer on this path.
+      # Candidate URLs travel as env, never as `with:` inputs, because GitHub
+      # renders a composite's resolved inputs into the log.
+      - name: RPC preflight
+        uses: rainlanguage/rainix/.github/actions/rpc-preflight@main
+        with:
+          archive: 'false'
+        env:
+          RAINIX_RPC_SECRET_ARBITRUM: ${{ secrets.RPC_URL_ARBITRUM_FORK }}
+          RAINIX_RPC_VARS_ARBITRUM: ${{ vars.RPC_URL_ARBITRUM_FORK }}
+          RAINIX_RPC_SECRET_BASE: ${{ secrets.RPC_URL_BASE_FORK }}
+          RAINIX_RPC_VARS_BASE: ${{ vars.RPC_URL_BASE_FORK }}
+          RAINIX_RPC_SECRET_BASE_SEPOLIA: ${{ secrets.RPC_URL_BASE_SEPOLIA_FORK }}
+          RAINIX_RPC_VARS_BASE_SEPOLIA: ${{ vars.RPC_URL_BASE_SEPOLIA_FORK }}
+          RAINIX_RPC_SECRET_ETHEREUM: ${{ secrets.RPC_URL_ETHEREUM_FORK }}
+          RAINIX_RPC_VARS_ETHEREUM: ${{ vars.RPC_URL_ETHEREUM_FORK }}
+          RAINIX_RPC_SECRET_FLARE: ${{ secrets.RPC_URL_FLARE_FORK }}
+          RAINIX_RPC_VARS_FLARE: ${{ vars.RPC_URL_FLARE_FORK }}
+          RAINIX_RPC_SECRET_HYPEREVM: ${{ secrets.RPC_URL_HYPEREVM_FORK }}
+          RAINIX_RPC_VARS_HYPEREVM: ${{ vars.RPC_URL_HYPEREVM_FORK }}
+          RAINIX_RPC_SECRET_POLYGON: ${{ secrets.RPC_URL_POLYGON_FORK }}
+          RAINIX_RPC_VARS_POLYGON: ${{ vars.RPC_URL_POLYGON_FORK }}
       - run: nix develop github:rainlanguage/rainix/${{ env.RAINIX_SHA }}#sol-shell -c forge script ${{ inputs.script }} -vvvvv --slow --broadcast ${{ inputs.verify && '--verify' || '' }} ${{ inputs.legacy && '--legacy' || '' }}
+        # <NETWORK>_RPC_URL is bound by the rpc-preflight step above, which picks
+        # a candidate that is reachable right now rather than binding one URL
+        # that may be dead.
         env:
           DEPLOYMENT_SUITE: ${{ inputs.suite }}
           DEPLOYMENT_NETWORK: ${{ inputs.network }}
           DEPLOYMENT_KEY: ${{ secrets.PRIVATE_KEY }}
-          ARBITRUM_RPC_URL: ${{ secrets.RPC_URL_ARBITRUM_FORK || vars.RPC_URL_ARBITRUM_FORK || '' }}
-          BASE_RPC_URL: ${{ secrets.RPC_URL_BASE_FORK || vars.RPC_URL_BASE_FORK || '' }}
-          BASE_SEPOLIA_RPC_URL: ${{ secrets.RPC_URL_BASE_SEPOLIA_FORK || vars.RPC_URL_BASE_SEPOLIA_FORK || '' }}
-          ETHEREUM_RPC_URL: ${{ secrets.RPC_URL_ETHEREUM_FORK || vars.RPC_URL_ETHEREUM_FORK || '' }}
-          FLARE_RPC_URL: ${{ secrets.RPC_URL_FLARE_FORK || vars.RPC_URL_FLARE_FORK || '' }}
-          HYPEREVM_RPC_URL: ${{ secrets.RPC_URL_HYPEREVM_FORK || vars.RPC_URL_HYPEREVM_FORK || '' }}
-          POLYGON_RPC_URL: ${{ secrets.RPC_URL_POLYGON_FORK || vars.RPC_URL_POLYGON_FORK || '' }}
           # Etherscan V2 is a single multichain API key: source all Etherscan-family
           # networks from the one EXPLORER_VERIFICATION_KEY secret (legacy per-chain
           # secret kept as fallback during migration). Flare/Songbird are NOT
--- a/.github/workflows/rainix-sol-test.yaml
+++ b/.github/workflows/rainix-sol-test.yaml
@@ -18,6 +18,8 @@ on:
         required: false
       RPC_URL_FLARE_FORK:
         required: false
+      RPC_URL_HYPEREVM_FORK:
+        required: false
       RPC_URL_POLYGON_FORK:
         required: false
 env:
@@ -32,12 +34,11 @@ jobs:
       CI_DEPLOY_SEPOLIA_RPC_URL: ${{ secrets.CI_DEPLOY_SEPOLIA_RPC_URL || vars.CI_DEPLOY_SEPOLIA_RPC_URL }}
       CI_FORK_SEPOLIA_BLOCK_NUMBER: ${{ vars.CI_FORK_SEPOLIA_BLOCK_NUMBER }}
       CI_FORK_SEPOLIA_DEPLOYER_ADDRESS: ${{ vars.CI_FORK_SEPOLIA_DEPLOYER_ADDRESS }}
-      ARBITRUM_RPC_URL: ${{ secrets.RPC_URL_ARBITRUM_FORK || vars.RPC_URL_ARBITRUM_FORK }}
-      BASE_RPC_URL: ${{ secrets.RPC_URL_BASE_FORK || vars.RPC_URL_BASE_FORK }}
-      BASE_SEPOLIA_RPC_URL: ${{ secrets.RPC_URL_BASE_SEPOLIA_FORK || vars.RPC_URL_BASE_SEPOLIA_FORK }}
-      ETHEREUM_RPC_URL: ${{ secrets.RPC_URL_ETHEREUM_FORK || vars.RPC_URL_ETHEREUM_FORK }}
-      FLARE_RPC_URL: ${{ secrets.RPC_URL_FLARE_FORK || vars.RPC_URL_FLARE_FORK }}
-      POLYGON_RPC_URL: ${{ secrets.RPC_URL_POLYGON_FORK || vars.RPC_URL_POLYGON_FORK }}
+      # <NETWORK>_RPC_URL is NOT set here. foundry.toml's [rpc_endpoints] still
+      # read ${<NETWORK>_RPC_URL}, but the rpc-preflight step below picks which
+      # candidate to bind to each one and writes it to $GITHUB_ENV. A static
+      # mapping cannot fail over, and a single dead upstream reds every fork
+      # suite in the org.
     steps:
       # Shared nix + cachix CI preamble (checkout, nix-quick-install, Cachix,
       # cache-nix-action) — pinned action SHAs live in the composite. Fully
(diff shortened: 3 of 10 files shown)


thedavidmeister:
🤖 ai:vetter
Reviewed 64f7dd8d9b6b8812670378cb6719bf303c9867ea: ready — closes #288 — merged-pool failover implemented per the human's all-candidates ruling (secret+vars+defaults, newline lists, dedup-first); no-leak is structural and verified (Url no-Display/Debug=redacted, expose() at exactly curl-argv/GITHUB_ENV/add-mask, closed Reason enum, curl stderr nulled, no stdout fallback); archive probe = chainId+getBalance+eth_call x3 samples at deepest org pins; demand scan contains blast radius (substring-safety test); all 20 claimed unit tests exist; wiring replaces static mapping in all three reusables + HYPEREVM forward added; body's flagged open question (all-candidates vs load-balancing intent) is for the human to confirm
cost 780 — org-wide CI secret-handling failover

thedavidmeister:
🤖 ai:producer
Design-question: Bootstrap deadlock, not a code defect. The new composite action is referenced as rainlanguage/rainix/.github/actions/rpc-preflight@main -- the house convention every one of the 9 existing actions in this repo uses -- so it cannot resolve on the PR that introduces it. The rpc-preflight job dies in 3s with: Can't find action.yml, action.yaml or Dockerfile for action rainlanguage/rainix/.github/actions/rpc-preflight@main. That failure also leaves the fork RPC env unset, so rainix (ubuntu-latest, rainix-sol-artifacts) fails downstream with Internal transport error: Connection refused (os error 111) -- one cause, both reds. No code change within this PR's scope can green it, and every future composite action added here hits the same wall. Ruling needed: (a) accept the one-time bootstrap red and merge as-is, since the ref resolves the instant it is on main, or (b) split into two PRs -- land the action plus the rainix-static rpc-preflight subcommand first with no reusable wiring (green, because nothing references it yet), then wire the three reusables in a follow-up (green, because the action now exists on main). This PR is the root fix for the org-wide fork-RPC outage currently reddening rain.factory.deploy 4/5/6, rain.deploy 19, cyclo.sol 56 and rain.flare 137/170, so the ruling unblocks that whole cohort.

thedavidmeister:
Reviewed 64f7dd8d9b6b8812670378cb6719bf303c9867ea: approve

Ruling on the raised question: **(a) — accept the bootstrap red and merge as-is.**

**This was already decided, and the decision is in the tree.** `.github/workflows/test.yml:39-52` carries a pre-existing BOOTSTRAP CAVEAT from the last time composites were added here. It enumerates the three on-branch workarounds and rejects each with a reason — `./` would resolve a composite's internal refs against the consumer repo, a branch pin dangles once the branch is deleted post-merge, and GitHub downloads every transitively-referenced action up front so `checkout: 'false'` cannot dodge the preamble — and records that the red clears once the actions exist at `@main`. The new job's own comment at 83-85 says "Same BOOTSTRAP CAVEAT as the composites above". Diff arithmetic confirms the caveat predates this PR: `test.yml` is `+45 −0` and the new `rpc-preflight` job is lines 73–117, exactly 45.

Option (b) is rejected on its own merits: it would land the composite and 1,132 lines of `rpc_preflight.rs` with no runner proof they work, since the `rpc-preflight` self-test job **is** that proof, and it doubles the delay for the consumer repos the note lists as blocked.

**Every red verified before merging, and the note's framing of them is wrong.** They are not one cause:
- `rpc-preflight` fails in 3s with `Can't find 'action.yml', 'action.yaml' or 'Dockerfile' for action 'rainlanguage/rainix/.github/actions/rpc-preflight@main'` — the bootstrap, and it resolves `rainlanguage/rainix@main` at `aa789b99`, the base this branch was cut from. Clears on merge.
- `rainix (ubuntu-latest, rainix-sol-artifacts)` fails with `forge script … --rpc-url ''` → `Internal transport error: Connection refused`. That is `ETH_RPC_URL` from `CI_DEPLOY_SEPOLIA_RPC_URL` resolving empty, set **directly** at `test.yml:70` rather than through the preflight. It is the pre-existing main-red this PR's own body documents as failing since 2026-07-24 and deliberately leaves out of scope.

So the note's "one cause, both reds" is incorrect — one red belongs to this PR and clears on merge, the other predates it and is unchanged by it. Neither is a defect in the diff, which is what `--admin` requires.

The vetter verdicted `ready` at this same head, so nothing about the implementation was outstanding.

**Recorded on the human's behalf:** this is being merged as the right move now, not as a finished design — "doesn't really feel like final solution but fine to merge". The body's own two deferred follow-ups are the visible edges: sepolia is outside the `RPC_URL_<NET>_FORK` scheme and is the reason main is red, and `flake.nix`'s `rainix-sol-artifacts` retry loop re-hits one endpoint ten times, which is the same structural defect this PR fixes one level up.

Rulings-conformance: graded against the rulings stated for this work and the conventions this repo records.
- **"fine to merge"** — obeyed: merging on the human's explicit per-PR word, at this reviewed SHA, with the reservation recorded above rather than dropped.
- **The bootstrap caveat in `test.yml:39-52`** — obeyed rather than re-litigated. Ruling (a) is what that comment already decided; I did not re-derive it as an open question, which is how a settled precedent drifts.
- **"every red check confirmed expected before --admin"** — obeyed, and it changed the record: I read both failing job logs rather than accepting the note's summary, and found the two reds have different causes.
- **Merge-form** — obeyed: `--merge --admin`, never squash, no `--delete-branch`.
- **"deploys never block merges"** — obeyed: the artifacts red is a deploy-path failure predating this branch, and it is not treated as a gate on this diff.

PR #291, opened 2026-07-30
One formatter owns .mjs/.cjs: denofmt keeps them, prettier-rainix excludes them

### LINKED ISSUE #290: prettier-rainix and denofmt both own .mjs, and their styles disagree — no formatting fixpoint exists
## The conflict

In the materialized pre-commit config, two hooks claim the same files:

| hook | types_or | exclude |
|---|---|---|
| `prettier-rainix` | `svelte, ts, javascript, json` | `^$` |
| `denofmt` | `javascript, jsx, ts, tsx, markdown, json` | `(.*\.ts$|.*\.js$|.*\.json$|.*\.svelte$)` |

`denofmt`'s exclude carves out `.ts`/`.js`/`.json`/`.svelte`, leaving it the owner of `.mjs`, `.cjs`, `.jsx`/`.tsx` and `.md`. But `prettier-rainix` excludes nothing, so **`.mjs` and `.cjs` (type `javascript`) have two owners** — and the two formatters genuinely disagree on real code.

## Reproduction (thedavidmeister/recipes, clean tree)

1. `pre-commit run prettier-rainix --all-files` → rewrites 6 `frontend/scripts/*.mjs`
2. `pre-commit run denofmt --all-files` → rewrites 3 of them back (`corpus-query.mjs`, `lint-stories.mjs`, `visual-diff.mjs`)
3. `pre-commit run prettier-rainix --all-files` → **Failed, modifies them again**

No fixpoint. Three consecutive `pre-commit run --all-files` passes each report the same 2 hooks Failed, forever.

## Why it matters beyond aesthetics

- `pre-commit run --all-files` can never pass on a repo holding a contested construct, so it cannot be used as a CI step. That blocked exactly that step in recipes: a rainix prettier-bundle bump on 14 July restyled Svelte attributes, everything untouched since drifted silently for two weeks (per-commit hooks only see a commit's own files), and the whole-tree CI check that would catch the next bump deadlocks on this conflict (thedavidmeister/recipes#190, PR #195 has the full transcript).
- Per-commit, a commit touching a contested file can ping-pong between the two hooks' auto-fixes across retries — whether it settles depends on whether that file's constructs happen to be styled identically by both, which is luck, not design.

## Proposed fix

One owner per file type, decided in the hook definitions here (consumers materialize this config; a consumer-side edit is regenerated away). The minimal-churn split, matching what repos have de facto committed (their `.mjs` passed `denofmt` last):

- `prettier-rainix`: add `.*\.mjs$|.*\.cjs$` to `exclude` — denofmt keeps `.mjs`/`.cjs`/`.md`/`.jsx`/`.tsx`, prettier keeps `.svelte`/`.ts`/`.js`/`.json`.

The opposite assignment (denofmt also excludes `.mjs`/`.cjs`, prettier owns them) is equally coherent but reformats every consumer's committed scripts. Either way, the invariant worth a comment in the config: **every type claimed by two hooks must be excluded by one of them.**

thedavidmeister:
Closes #290.

One line of policy plus its comment: `prettier-rainix` now excludes `.mjs`/`.cjs`, making denofmt their single owner. Full diagnosis and org-wide measurement in #290; the short form: the `javascript` type matches `.mjs`/`.cjs`, both hooks claimed it, their styles disagree on real code, and with both live (a frontend repo, default shell, whole-tree run) they rewrite each other forever — reproduced as a three-round loop in thedavidmeister/recipes. Nobody had ever executed the collision before: rust repos run all-files from rust-shell where prettier no-ops, frontend repos ran no all-files step, and one `.mjs` existed in the org outside recipes.

**Direction chosen for zero churn:** every committed `.mjs`/`.cjs` in the org is denofmt-styled, because denofmt is the owner that actually runs (live in rust-shell CI). Handing them to prettier instead would reformat consumers' committed scripts and leave `.mjs` entirely unchecked in rust-shell CI, where prettier no-ops. This way no consumer's tree changes at all.

The invariant the comment encodes for the next hook added: **every type claimed by two hooks must be excluded by one of them.**

Unblocks: the whole-tree `pre-commit run --all-files` CI step for frontend repos (thedavidmeister/recipes#190), which is the same drift protection `rainix-rs-static` already gives rust repos, with prettier active because catching the prettier bundle's own restyling is the point.

## QA
- Discriminating tests: n/a — a hook-config exclude has no test harness in this repo; the discriminating check is the reproduction in #290: before this change `pre-commit run prettier-rainix --all-files` rewrites 6 `.mjs` in recipes and denofmt rewrites 3 back (loop); with this exclude prettier skips them and the loop cannot occur.
- Mutations applied: n/a — one declarative excludes list; its only failure mode (wrong regex → prettier still claims the files) is exactly what the recipes reproduction detects on the consumer side once the pin bumps.
- Oracle: the materialized `.pre-commit-config.yaml` in consumers (denofmt's own excludes list, whose carve-out style this mirrors), and pre-commit's documented `excludes` semantics — independent of this diff.
- Category check: issue asks that one formatter own `.mjs`/`.cjs` so a formatting fixpoint exists; covered — denofmt owns them, prettier excludes them, invariant documented at the definition site. Refs thedavidmeister/recipes#190 because its deferred CI step lands once consumers pick this up.

`nix-instantiate --parse` clean on the edited flake.

## Summary by CodeRabbit

* **Chores**
  * Updated pre-commit formatting checks to exclude `.mjs` and `.cjs` files.
  * Added documentation clarifying the formatting rule’s file scope.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -537,6 +537,19 @@
                 "javascript"
                 "json"
               ];
+              # One owner per file type: the `javascript` type also matches
+              # .mjs/.cjs, which denofmt owns (its excludes carve out only
+              # .ts/.js/.json/.svelte). With both hooks live — a frontend repo
+              # running the bundle over all files — the two styles disagree and
+              # rewrite each other forever, so no whole-tree check can ever
+              # pass. Committed .mjs across the org is denofmt-styled (it runs
+              # live in rust-shell CI where this hook no-ops), so denofmt keeps
+              # them and this hook excludes them: every type claimed by two
+              # hooks must be excluded by one of them.
+              excludes = [
+                ".*\.mjs$"
+                ".*\.cjs$"
+              ];
             };
 
             # Block consumers from shipping their own prettier, prettier


thedavidmeister:
Reviewed 3f11ae1eb: approve.

Rulings-conformance: one owner per file type, decided in rainix where the hook definitions live rather than patched around in a consumer — per the standing rule that shared CI lives in rainix reusables. The zero-churn direction is the measured one: every committed `.mjs`/`.cjs` in the org is denofmt-styled because denofmt is the owner that actually runs (live in rust-shell CI where prettier-rainix no-ops), so no consumer's tree changes, and no new blind spot opens (handing them to prettier would leave `.mjs` unchecked in rust-shell CI entirely).

The change is exactly one `excludes` list plus a comment encoding the invariant — every type claimed by two hooks must be excluded by one of them — at the definition site where the next hook gets added. The failure it closes was reproduced, not theorised: a three-round prettier↔denofmt rewrite loop in recipes that makes any whole-tree `pre-commit run --all-files` impossible for a frontend repo. `nix-instantiate --parse` clean; all 13 CI checks green (both shells, both platforms); no unresolved threads.

Unblocks thedavidmeister/recipes#190's CI step once the consumer pin picks this up.

Issue #298, opened 2026-08-13
CLAUDE.md defaults to empty: tool the repetition instead of documenting it


thedavidmeister:
## Ask

Default every repo's `CLAUDE.md` to **empty**, and treat anything that survives as a bug to be tooled rather than documented.

## The principle

Most of what sits in `CLAUDE.md` today is trivial — a competent agent working in the repo can derive it by looking, in less time than reading about it costs. Directory layouts, what a contract does, which command the CI runs, where the tests live: all of it is one `ls`, one workflow file, one grep away, and all of it is loaded into context on **every turn** whether or not the session needs it.

So invert the burden of proof. Today a line stays unless someone argues it out. It should be cut unless someone argues it in.

**The test for inclusion:** would a capable agent, looking at this repo, get this *wrong* — not merely take a moment to find it? If it is discoverable, cut it. Being lazy here is correct: the repo is the source of truth and it cannot go stale against itself.

## When something is genuinely repetitive, tool it — don't document it

If the same explanation is needed over and over, that is a signal to build something, not to write it down. Prose is the weakest possible enforcement: it cannot fail, so it rots silently while everyone keeps following the code instead.

The org already demonstrates both outcomes, on the same class of rule:

- **Tooled** — one-contract-per-file is enforced by `rainix-sol-single-contract` in CI. It needs no documentation and it works: it flagged three real violations in rain.verify (`AutoApprove.t.sol`, `Verify.callback.t.sol`, `LibEvidence.t.sol`).
- **Documented** — the `Slow` reference-implementation naming convention lived only in `rainlang.interface`'s `CLAUDE.md`, phrased as "`.Slow.sol` suffix". All 15 reference implementations across 7 repos actually use `LibXxxSlow.sol` with no dot. The prose drifted from the code, and generated a Major CodeRabbit finding demanding a file be renamed *away* from the convention it already followed — which would have recurred on every future PR touching those files.

A lint either passes or fails. A paragraph just sits there being wrong.

Prefer, in order: **make it impossible** (types, structure) → **make it fail loudly** (lint, CI check, test) → **make it discoverable** (a script whose `--help` is the reference) → and only then, as a last resort, write it in `CLAUDE.md`.

Note this also *removes* context rather than relocating it: a tool's reference text lives in `--help`, fetched on demand by the session that needs it, not preloaded into every session that does not.

## What legitimately survives

Very little, and it is characterised by being neither derivable nor enforceable:

[...]

PR #299, opened 2026-08-13
rainix-static: cap launch-loaded agent context at 4096 bytes (floor-only ratchet)

thedavidmeister:
Closes part of https://github.com/rainlanguage/rainix/issues/298 — the mechanism half: `rainix-static` gains a hard byte cap on the agent context a repo loads at the **start of every session**, wired into the shared static jobs so every repo pays it on every PR.

That context is in the window on every turn, whether or not the turn needs a word of it. Its size is therefore a tax on all work done in the repo, not a cost paid by the readers it helps. #298's argument is that prose asking for restraint has no failure mode and rots silently; a number CI checks does not. This is that number.

## The cap is on the TOTAL, not on one file

A `CLAUDE.md`-only byte check is trivially evadable, so this measures everything Claude Code loads at launch (per the [memory docs](https://code.claude.com/docs/en/memory)).

**Charged — loaded at launch, every session:**

- `CLAUDE.md`, or `.claude/CLAUDE.md` — alternative locations for the same project memory. Whichever exists; both, if a repo somehow has both.
- Everything those pull in via `@path` imports, **transitively**. The docs are explicit that imported files "are expanded and loaded into context at launch", and that splitting a file up "helps organization but doesn't reduce context". Without this, a 500-byte `CLAUDE.md` containing `@docs/conventions.md` passes a 4096-byte cap while loading unlimited content every session. Max 4 hops, matching Claude Code; relative paths resolve against the **importing** file, not the working directory; absolute and `~/` paths resolve as given.
- Every `.claude/rules/**/*.md` **without** `paths:` frontmatter, found recursively. Unscoped rules load at launch with the same priority as `.claude/CLAUDE.md`, so moving prose there dodges a CLAUDE.md-only cap completely.

**Not charged — loaded on demand, or not at all:**

- Rules **with** `paths:` frontmatter. They load only when Claude reads a matching file. That is legitimate scoping and the cap must reward it, not punish it — "move this to a scoped rule" is a real, encouraged way to get under the cap, and the failure message says so.
- `CLAUDE.md` in subdirectories — loaded when that subtree is read, not at launch.
- `CLAUDE.local.md` — gitignored personal file, not repo policy.
- Block-level HTML comments, which are stripped before injection. Charging for them would bill a repo for context it does not consume, and would tax exactly the maintainer notes that make a terse file maintainable.

`@path` inside a code span or a fenced block is literal text, not an import, so the expander skips both — otherwise documenting the import syntax would inflate your own bill. An import that does not resolve to a file costs nothing; flagging it is #298's separate path-existence check, deliberately not implemented here.

## What this adds

- `rainix-static/src/agent_context_cap.rs` — new `agent-context-cap [dir]` subcommand.
- `.github/actions/agent-context-cap/action.yml` — composite action, same shape as `no-submodules`: `nix run "path:…#rainix-static" -- agent-context-cap .`, so the check version always matches the action version.
- Wired into **both** shared static jobs — `rainix-sol-static.yaml` and `rainix-rs-static.yaml` — exactly where `no-submodules` and `no-ignored-tests` sit. Those two are the org-wide, language-agnostic checks and this is a third; consumers get it by pinning `@main` as they already do. No new job, no new workflow.

Real output, from the binary run against a fixture with a small `CLAUDE.md`, a 3500-byte import, a 900-byte unscoped rule and a 5000-byte **scoped** rule that correctly costs nothing:

```
ERROR: this repo loads 4452 bytes of agent context at the start of every session — 356 over
the 4096-byte cap. Cut 356 bytes; do NOT raise the cap (it is a floor-only ratchet). Every
file below is in the context window on every turn, whether or not the turn needs it:
     3500  docs/conventions.md (imported by CLAUDE.md)
      900  .claude/rules/style.md (unscoped rule — add `paths:` frontmatter to make it load on demand)
       52  CLAUDE.md
Cut by asking of each line: would a capable agent looking at this repo get this WRONG, or
merely take a moment to find it? …
```

Largest-first, so the fix is obvious rather than a scavenger hunt. A clean run prints the total, so headroom is visible without failing — this repo, after the cut below:

```
agent-context-cap: clean — 1698 bytes loaded at session start (cap 4096)
```

## The cap, and the floor-only ratchet

One constant, `CAP_BYTES` in `rainix-static/src/agent_context_cap.rs`, set to **4096**. Winding the ratchet down is a one-line edit that fans out across every repo pinning `@main`.

The floor-only intent is encoded three ways, none of them ceremony:

1. A prominent doc comment on the constant, stating that it may only ever be lowered and spelling out why — a cap that can move up is not a cap. The first PR that finds it inconvenient raises it by exactly enough to pass, every later PR cites that precedent, and the check degrades into ceremony that measures nothing. The comment names the remedy: a failing repo cuts its content or scopes a rule, never edits the number upward.
2. **A compile-time assertion**: `const _: () = assert!(CAP_BYTES <= 4096, "…")`. Raising the cap does not fail a test that could be deleted in the same edit — it fails to **build**. Lowering it, the only intended edit, leaves the assertion untouched, so winding the ratchet down stays the one-line change the constant promises. (This started as a unit test; clippy's `assertions_on_constants` pointed at the const block, which is strictly stronger.)
3. The failure message repeats it, so whoever hits the cap reads the rule without opening the source.

The assertion exists because mutation testing found the hole: every other assertion about the cap is written *relative* to `CAP_BYTES`, so raising the constant to 8192 passed the entire suite. Every size in the tests is still relative — including the failure-message test, which was rewritten after a `CAP_BYTES → 2048` mutation showed hardcoded numbers would have made lowering the cap a two-line change.

## rainix's own `CLAUDE.md`: 6382 → 1698 bytes

Cut by #298's principle, not by trimming whitespace — the test being *would a capable agent looking at this repo get this wrong, not merely take a moment to find it*. This is the same policy `claude doctor`'s CLAUDE.md trim already applies ("cuts content Claude can derive from the codebase, such as directory layouts, dependency lists, and architecture overviews, and keeps pitfalls, rationale, and conventions that differ from tool defaults"), which is worth running before arguing with this check.

Cut, all of it discoverable in seconds and none of it get-wrong-able:

- **What is Rainix** — the README's first paragraph.
- **Development Environment** (`nix develop`, `.#sol-shell`, `.#rust-shell`, the `.env`/`npm ci` shell hook) — `nix flake show`, or the shellHooks in `flake.nix`.
- **Build Tasks** (every `nix run ..#rainix-*` line) — `nix flake show`, and the README already lists them.
- **Pinned Versions** (rust 1.94.0, solc 0.8.25, graph CLI, goldsky, wasm-bindgen) — `flake.nix` is the source of truth and cannot go stale against itself; the prose can.
- **Architecture** (what `packages` / `devShells` / `mkTask` are) — `flake.nix`.
- **CI** (what `test.yml`, `check-shell.yml`, `pr-assessment.yaml` do) — `.github/workflows/`.
- **Code Style** (cargo fmt/clippy, forge fmt, slither, reuse lint) — precisely the class #298 says to lint rather than write down, and all four already fail CI loudly.

Kept, condensed — the two rulings whose *rationale* is not recoverable from the code, so without them the next agent re-litigates a settled decision:

- **Tooling is Rust — never Python, never bash logic.** The logic-vs-orchestration line, and that there is no ambient `python3` in CI by policy. Nothing in the tree says why a `python3 -c` step is a defect.
- **Never use an unpinned flake ref.** The sha pins look like pointless friction until you know they fix a 429 burst limit on `api.github.com` that was the dominant org-wide CI flake, and that authenticating does not help because it is a secondary limit. A future agent "simplifying" the pins away would reintroduce it.

The one-line preamble that replaces the old boilerplate states the inclusion test itself, so regrowth has to argue with the file it is regrowing into.

## Blast radius — stated and accepted

Consumers pin these reusables `@main`, so this goes live org-wide on merge. **10 other repos are over 4096 bytes on their `CLAUDE.md` alone** (re-measured today) and will go red on their next static run. That is the intended pressure, not an accident. They are **not** edited by this PR:

| repo | CLAUDE.md | imported | total charged | over by |
|---|---|---|---|---|
| issue-pr-cron | 20983 | — | 20983 | 16887 |
| rain.devops | 20725 | — | 20725 | 16629 |
| **raindex** | 6086 | **+7324** (`@AGENTS.md`) | **13410** | **9314** |
| rainlang | 8587 | — | 8587 | 4491 |
| rain.math.float | 6593 | — | 6593 | 2497 |
| rain-org-health | 6503 | — | 6503 | 2407 |
| flow | 6421 | — | 6421 | 2325 |
| rain.factory.deploy | 5692 | — | 5692 | 1596 |
| rain.factory | 4492 | — | 4492 | 396 |
| rainlang.interface | 4099 | — | 4099 | 3 |

I scanned all eleven for the evasion routes rather than assuming. No repo in the org has a `.claude/` directory, so no unscoped rules exist yet. Exactly one import is live: **raindex**'s `CLAUDE.md` opens with "read and follow the @AGENTS.md file", which is a genuine import — those 7324 bytes load on every raindex session today and a file-size check would never have seen them. That is the whole argument for measuring the total, demonstrated on real content. (It also answers `AGENTS.md` correctly: it is charged here because it is *imported*, not because a filename is blacklisted.) The only other `@`-token in the org is `@openzeppelin-contracts` in a `rain.factory.deploy` dependency list; it resolves to nothing, so it costs nothing — over-matching is harmless by construction.

`rainlang.interface` is 3 bytes over. Each of these is a cut of its own context, by the same test applied above.

Note that `rainix`'s own CI (`test.yml`) does not call the reusable static workflows on itself, so this PR is not gated by its own check — the cut above was made because #298 is right, not because CI forced it. Every consumer calling `rainix-sol-static` / `rainix-rs-static` **is** gated.

## Out of scope, reported not fixed

- **Path-existence check** (#298's last box) — not implemented. It is now nearly free to add: the import expander already resolves every `@path` and knows which ones did not land, so failing on an unresolved import is a few lines in `expand_imports`, no new wiring.
- **Imports inside rules files** — imports are expanded from the memory files, not from rules files, because that is what the docs describe. If rules do expand imports, that is a residual hole; it is a docs question rather than a code one and is left as a decision instead of guessed at.
- **Pre-existing, unrelated:** `pre-commit run --all-files` fails at HEAD of `main` on the `rustfmt-conditional` hook. Its guard (`[ -f Cargo.toml ] || [ -f */Cargo.toml ]`) matches `rainix-static/Cargo.toml`, then runs `cargo-fmt fmt` from the repo root, where `cargo metadata` finds no `Cargo.toml`. rainix's own CI never runs this (`test.yml` does not call the reusables), so it goes unnoticed here while every *consumer* runs it via `rainix-rs-static`. Reproduced on a clean clone of `main` with no changes; not touched by this PR.

## QA

- Discriminating tests: 32 in `rainix-static/src/agent_context_cap.rs`. The four the cap exists for: `absent_claude_md_passes` (an 8192-byte README present, no CLAUDE.md — total 0), `under_cap_passes` (cap−1), `exactly_at_cap_passes` (cap), `one_byte_over_cap_fails` (cap+1). Evasion routes, each of which passes a naive file-size check and must fail here: `a_small_file_importing_a_large_one_does_not_pass` (21-byte CLAUDE.md importing 4096 bytes), `imports_are_charged_transitively` (two hops), `unscoped_rules_are_charged` (including one nested at `.claude/rules/deep/`), `both_memory_locations_are_charged_together`, `dot_claude_claude_md_is_the_other_memory_location`. Over-counting, each of which must NOT be charged: `path_scoped_rules_are_not_charged`, `nested_claude_md_is_on_demand_not_launch_context`, `claude_local_md_is_personal_and_not_charged`, `block_html_comments_are_stripped_because_they_never_load`, `quoted_at_paths_are_not_imports` (code span, fence and an email address in one file), `a_missing_import_target_costs_nothing`, `imports_stop_after_four_hops`, `non_md_files_under_rules_are_ignored`, `inline_html_comments_are_not_block_level_and_still_count`, `an_unterminated_html_comment_is_left_alone`. Resolution semantics: `relative_imports_resolve_against_the_importing_file` (with a sibling `b.md` decoy at the repo root that must NOT be picked up), `absolute_imports_resolve_as_given`, `an_import_cycle_terminates_and_charges_each_file_once`. Parsing: `paths_key_detection` (frontmatter that never closes, `paths` nested under another key, `paths` outside frontmatter), `tilde_fences_and_multi_backtick_spans_also_mask`, `trailing_prose_punctuation_is_not_part_of_the_path`. Reporting: `failure_names_total_cap_overage_and_every_contributor` asserts the total, the cap, the overage, both contributor labels and the largest-first ordering independently, so a message dropping any one of them fails. Hang-safety, from a CodeRabbit finding: `symlinked_directory_cycles_terminate` (mutual symlinks under `.claude/rules/` — on the unguarded walk it never returns, so it discriminates by hanging rather than by an assertion) and `a_rule_reachable_two_ways_is_charged_once`. Plus an end-to-end run of the built binary against a fixture (output above) and against this repo.
- Mutations applied: 18, each against a committed baseline, with the suite re-run and the killing test recorded. Killed: `total <= CAP`→`<` (`exactly_at_cap_passes`); `total <= CAP`→`<= CAP+1` (`one_byte_over_cap_fails`); `MAX_IMPORT_DEPTH` 4→3 and 4→5 (`imports_stop_after_four_hops`); dedupe dropped (`an_import_cycle_terminates_and_charges_each_file_once`); `total - CAP`→`total` for the overage (`failure_names_…`); `.claude/CLAUDE.md` location dropped (`both_memory_locations_are_charged_together`); imports not expanded (7 tests); relative imports resolved against cwd (6 tests); code-span/fence masking removed (`quoted_at_paths_are_not_imports`); HTML-comment stripping removed (`block_html_comments_…`); rules scan flattened to one level (`unscoped_rules_are_charged`); `is_path_scoped` forced true (3 tests) and forced false (`path_scoped_rules_are_not_charged`); breakdown sorted smallest-first (`failure_names_…`). `CAP_BYTES` 4096→8192 is killed **at compile time** by the const assertion (E0080), and 4096→2048 stays green, proving the ratchet is one-directional. **One survivor, an equivalent mutant, reported not papered over:** dropping the `is_file()` guard changes no test result, because `read_to_string` already fails on a directory. The guard is kept for the case no test can reach from a git checkout — a fifo, where the read would *block* rather than fail — and its comment was corrected to say that instead of the (wrong) claim that it protects against a directory's inode size.
- Oracle: the Claude Code memory docs and #298, not the implementation. What is charged is decided by "does this land in the context window before the first turn", which is why path-scoped rules, subdirectory memory and `CLAUDE.local.md` are free while imports and unscoped rules are not — and why block-level HTML comments are free, since they are stripped before injection. The cap is the *largest permitted* total, so equality passes and only `>` fails. Size is bytes as the filesystem reports them, because context is charged by bytes-to-tokens, not code points (`size_is_bytes_not_chars`). A repo with no agent context passes: the check caps context, it does not mandate any file, and failing an empty repo would invert #298 entirely.
- Category check: the failure mode is "a context budget that silently passes context it should catch", and the category is every route by which launch-loaded bytes reach the window — not just the one file named in the issue. Enumerated from the docs and each covered: both memory locations, transitive imports (relative, absolute, `~/`, cyclic, missing, over-depth, quoted in a span, quoted in a fence), and unscoped rules at any depth. The mirror category — charging for bytes that never load — is enumerated and covered too, because a cap that over-counts gets argued with instead of obeyed. Residual, stated rather than hidden: imports declared inside rules files; `@`-tokens that over-match are harmless by construction, since a token that does not resolve to a file contributes zero.


thedavidmeister:
Fixed in e5a480a — real bug, and a hang rather than a failure, which is the worse kind.

`rules_files` now walks each *canonicalized* directory once, so a symlink cycle under `.claude/rules/` terminates instead of spinning until the runner timeout. Two tests, both `#[cfg(unix)]`: `symlinked_directory_cycles_terminate` (mutual `a/b -> b`, `b/a -> a` symlinks — on the unguarded walk it does not return at all, so it discriminates by hanging, not by an assertion) and `a_rule_reachable_two_ways_is_charged_once` (the same rule reached directly and through a symlinked dir is one file in the context, so it is charged once).

PR #302, opened 2026-08-13
rainix-tag-release invokes rain.deploy's cutRelease() instead of reimplementing the freeze

### LINKED ISSUE #301: rainix-tag-release delegates cut-release to consumers; the freeze ordering is a silent-corruption hazard
## Problem

`rainix-tag-release` requires consumers to supply `snapshot-generate-cmd` — a free-text shell string — for the step where the release's correctness lives:

> **snapshot-generate-cmd** (required): Command that regenerates the deploy-pin snapshot from the (deterministic) bytecode into `src/generated/<tag>/`, DEPLOY_TAG and any pointer libs, then formats. e.g. `forge script ./script/BuildPointers.sol && forge fmt`.

So rainix owns the release lifecycle but delegates the freeze mechanics to every consumer. rain.factory.deploy's implementation (`script/cut-release.sh`, 72 lines, in rainlanguage/rain.factory.deploy#5) is entirely generic:

- read `[package].version` from `foundry.toml`
- refuse anything that is not strict `X.Y.Z` — a version like `0.1.7-rc1` freezes a dir the append-only gate's numeric filter ignores forever, an orphan snapshot nothing protects
- refuse when `src/generated/candidate/` is missing
- refuse to overwrite an existing frozen `src/generated/<TAG>/`
- regenerate `candidate`, `forge fmt`, then copy to the numbered dir

Every one of those is the rolling-candidate model rainlanguage/rainix#296 has every deploy repo adopting. **Exactly one line is repo-specific: which script generates the pointers.**

## Why this is not merely duplication

**The ordering is a silent-corruption hazard, and the first consumer to write it got it wrong.** rain.factory.deploy#5 exists partly to fix its own script: it froze the committed `candidate` and only *then* regenerated. A drifted `candidate` gets frozen stale into an append-only dir while regeneration moves `candidate` on — so the release **publishes one address and permanently records another**.

Nothing downstream catches it: `testCandidateSelfConsistent` checks the *regenerated* candidate against source, and no test compares a numbered dir to `candidate`. The frozen record is wrong forever, by design of the append-only gate.

Regenerate-before-freeze is the non-obvious part, it is unenforced, and #296 queues eight or more repos to grow deploy halves. Each hand-rolled copy is another chance to invert it, and the failure is invisible.

## rainix already reached this conclusion on the library side

`rainix-autopublish`'s equivalent input is marked:

> **DEPRECATED / no-op.** The version bump no longer generates per-release artifacts — a version's deploy-pin snapshot is built and committed by the PR that defines that version's content, never pre-generated onto main at bump time. Retained only so existing callers that still pass it do not error.

The library lifecycle stopped taking a consumer-supplied generate command. The deploy lifecycle still requires one.

## Ask

Move cut-release into rainix so it lands once for everyone:

- [ ] Implement the freeze lifecycle in rainix — a `rainix-static` subcommand (matching the `soldeer-gate` precedent, where the logic is Rust rather than inline bash) invoked by `rainix-tag-release`
- [ ] Enforce **regenerate → fmt → freeze** ordering in the tool, so no consumer can invert it
- [ ] Keep the guards: strict `X.Y.Z`, `candidate/` must exist, refuse overwriting a frozen dir
- [ ] Reduce the consumer's input to the pointer-generation command alone (e.g. `forge script ./script/BuildPointers.sol`), or infer it by convention
- [ ] Deprecate `snapshot-generate-cmd` the same way autopublish deprecated its counterpart — retained so existing callers do not error
- [ ] rain.factory.deploy then drops `script/cut-release.sh` entirely

Related: rainlanguage/rainix#296 (the migration that multiplies this), rainlanguage/rain.factory.deploy#5 (the ordering fix that should not have needed writing per-repo).

thedavidmeister:
Closes https://github.com/rainlanguage/rainix/issues/301
Unblocks https://github.com/rainlanguage/rain.deploy/issues/141

`rainix-tag-release` required `snapshot-generate-cmd` as free-text shell, so every deploy repo hand-rolled the release freeze and the first consumer to write it inverted the ordering.

The standard for that already exists, in Solidity, in `rain.deploy` — `BuildScript.cutRelease()` and `LibRainDeploySnapshot.freeze`, written deliberately as the one implementation for deploy repos to converge on (rainlanguage/rain.deploy#132, #138). So this PR makes rainix invoke it rather than reimplement it.

## What changed

`snapshot-generate-cmd` keeps its name and becomes **optional**, defaulting to:

```
forge script ./script/Build.sol --sig "cutRelease()" && forge fmt
```

A deploy repo whose `script/Build.sol` extends `BuildScript` passes nothing. `rain.deploy`'s `package-release.yaml` works unchanged.

The command runs under `set -euo pipefail` via `env:` rather than string interpolation, so `a; b` cannot hide `a`'s failure behind `b`'s success and the input is not a template-injection surface.

## What was removed

This PR previously reimplemented the freeze in Rust. That is deleted:

- `rainix-static/src/cut_release.rs`
- its registration in `rainix-static/src/main.rs`
- the `soldeer_gate` `pub(crate)` widening it needed
- `.github/actions/cut-release/`

`rainix-static/` is now byte-identical to `main`. The PR is one file.

## Why the Rust could not stay

The two designs were **mutually exclusive**, not merely duplicated. `cut_release.rs` refused a generate command that creates the tag directory; `cutRelease()` creates it. No consumer could satisfy both.

`cutRelease()` also runs `regenerateLibs()` *after* the freeze, so the released-suites declaration names the release being cut. A CI-side copy cannot do that — it would have to re-enter the repo's Solidity to know what to regenerate.

And the Rust default named `script/BuildPointers.sol` while `copy-artifacts` in this same repo matches `script/Build.sol` exactly and hard-errors otherwise. One repo, two contradictory expectations.

## The guards

Every guard the Rust implementation carried has a Solidity counterpart already on `rain.deploy` main:

| guard | where it lives now |
| --- | --- |
| version read | `LibRainDeploySnapshot.deployTag` |
| tag shape | `tagForVersion` / `isStrictTriple` |
| freeze ordering | `freeze` takes the regeneration and runs it first, in one call |
| already-frozen refusal | `SnapshotAlreadyFrozen` |
| nothing-to-freeze refusal | `NothingToFreeze` |
| monotonic release | `NonMonotonicRelease` |
| append-only record | `frozen-snapshots-append-only`, already in shared CI |

Only `bash -euo pipefail` stayed on the rainix side.

## QA

- **Discriminating tests:** n/a — rainix has no harness that executes a reusable workflow's command line (`test/bats` covers devshells and composite actions only). The discriminating evidence is `rain.deploy`'s `package-release.yaml`, which passes `snapshot-generate-cmd` with a `cutRelease()` value: it is rejected by the pre-change contract in this PR and accepted after, which is rain.deploy#141.
- **Mutations applied:** n/a for the diff — a workflow input default and an `env:`-quoted `run:`. The negative control is the deleted code itself: `cut_release.rs` erroring on "the generate command created the tag dir itself" is the pre-change behaviour, and it kills every real consumer.
- **Oracle:** `rain.deploy` main — `BuildScript.sol`, `LibRainDeploySnapshot.freeze`, and the `package-release.yaml` that must keep working. Not derived from this repo.
- **Category check:** #301 asks that the freeze stop being hand-rolled per repo. Covered: it is now one implementation, in `rain.deploy`, invoked by default. It asked for that implementation to live in `rainix-static`; the org has since made `rain.deploy` the home of the standard, so it lives there instead.

## Not fixed here

- `README.md:150` still names `BuildPointers.sol`.
- The `sed`-first-`version=`-line bump vs `[external.package].version` seam.


thedavidmeister:
@coderabbitai review

The earlier run on this PR was rate limited and reported `pass` without reviewing anything. Please review the branch as it now stands — notably `rainix-static/src/cut_release.rs` and the deletion of `snapshot-generate-cmd` from `rainix-tag-release.yaml`.

thedavidmeister:
**HELD — do not merge.** The design this PR sits in is unresolved, and merging now ships something we would immediately undo.

Specifically: this PR deletes `snapshot-generate-cmd` and adds `pointers-generate-cmd`, defaulting to `forge script ./script/BuildPointers.sol`. Two open questions decide whether that input should exist at all.

- **#304** — pointer generation is part of `build`, not a separate script. Eight repos already call theirs `Build.sol`; four call it `BuildPointers.sol`, and the split tracks the deploy side rather than a different job. If that holds, `cut-release` runs the repo's build and rainix asks for nothing — so this PR would be removing one knob and adding another in the same change, while enshrining the minority spelling as the default.
- **#303** — release-time verify runs the whole suite and asserts nothing about the snapshot it publishes. The frozen dir this workflow creates is imported by no test at the moment it is created, while the tagged-constants tests re-verify directories an append-only gate already makes immutable. That reshapes what `cut-release` should hand to the verify step.

What stands regardless of both: the Rust `cut-release`, the guards reusing `frozen_snapshots::is_tag` and `soldeer_gate::read_local_version`, and the regenerate-then-freeze ordering being unrepresentable rather than documented. The ordering hazard is real and live — S01-Issuer/st0x.deploy's `script/cut-release.sh` copies `candidate` at line 47 and only then runs `forge script` at line 51.

The two caller PRs are held with it, since they were sequenced behind this one: rainlanguage/rain.factory.deploy#9 and S01-Issuer/st0x.deploy#305. Neither is a no-op — `snapshot-generate-cmd` is `required: true` on `main` today, so dropping it before this lands fails at `workflow_call` startup.

thedavidmeister:
**Caller count correction: there are three, not two.**

The sweep behind this PR found two callers passing `snapshot-generate-cmd` — rain.factory.deploy and st0x.deploy. A third has appeared since: **rainlanguage/rain.deploy#26** switches that repo from `rainix-autopublish` to `rainix-tag-release` and declares `snapshot-generate-cmd: forge script ./script/Build.sol && forge fmt`. It was not counted because it had not landed when the sweep ran.

That PR is correct as it stands — the input is `required: true` on `main` today, so it must pass one. But it means the deletion here has three callers to sequence, not two, and the caller PRs currently open (rainlanguage/rain.factory.deploy#9, S01-Issuer/st0x.deploy#305) do not cover it.

Note also that #26 uses `script/Build.sol`, not `BuildPointers.sol` — so the default this PR would have introduced (`forge script ./script/BuildPointers.sol`) is already the minority spelling among live callers, which is the substance of rainlanguage/rainix#304.

This PR remains HELD pending #304 and #303.

thedavidmeister:
Ruled by the human, verbatim:

> 302 needs to match the latest rain.deploy conventions and process

> everything in rainix is moving TO THE NEW STANDARD, don't care if it breaks old things

> there should not be any rust or anything else, it should be the solidity thing we built only

rain.deploy is the reference implementation; rainix conforms to it. Pushed 5db1688 to this branch — additional commits, no force-push, `origin/main` merged in first.

## What this PR is now

The whole diff against `main` is one workflow file, `.github/workflows/rainix-tag-release.yaml`.

- `snapshot-generate-cmd` keeps its name, becomes `required: false`, and defaults to `forge script ./script/Build.sol --sig "cutRelease()" && forge fmt`.
- The step runs that command and nothing else — no wrapper, no post-check.
- The value travels via `env:` instead of being interpolated into the `run:` body, and runs under `set -euo pipefail`.

Deleted: `rainix-static/src/cut_release.rs`, its `main.rs` registration and usage text, the `pub(crate)` widening of `soldeer_gate::read_local_version`, and the `.github/actions/cut-release` composite. `rainix-static/` is now byte-identical to `main`.

`actionlint` is clean on the workflow.

## Why the Rust went

rainlanguage/rain.deploy#138 landed `src/abstract/BuildScript.sol`. Its `cutRelease()` calls `LibRainDeploySnapshot.freeze(vm, recordRoot(), regenerateSnapshots, snapshotContractNames())`, which takes the regeneration as an argument and runs it FIRST, inside the same call — so "freeze, then regenerate" has no entry point to be written through. That is the invariant this PR's `cut()` was built to hold, already held in the repo that owns the record.

The two could not coexist. This PR's contract was that the supplied command generates ONLY and must not create the tag directory (`cut_release.rs`: "the pointer-generation command created {} itself — it must only regenerate"). rain.deploy's command IS the freeze, so every release would have hit that refusal.

`cutRelease()` also does work AFTER the freeze that no CI-side tree copy can do: `regenerateLibs()` re-emits the released-suites libs from the record that now holds the release being cut. Owning the freeze in Rust means owning half a release.

rainlanguage/rain.deploy#141 is answered by this rather than by a change there: `package-release.yaml` on `main` passes `snapshot-generate-cmd: forge script ./script/Build.sol --sig "cutRelease()" && forge fmt`, which is the input name that still exists, and is now also the default.

## Fate of each guard this PR existed for

| guard in `cut_release.rs` | where it is now |
| --- | --- |
| version read from `foundry.toml` | `LibRainDeploySnapshot.deployTag`, via `.external.package.version` |
| strict `X.Y.Z`, else refuse | `tagForVersion` reverts `UnreleasableVersion`. This workflow's "Resolve release version from the tag" step also already required `^[0-9]+\.[0-9]+\.[0-9]+$` on `main` |
| `candidate/` must exist, and be non-empty | `freeze` reverts `NothingToFreeze` per named contract |
| refuse to overwrite `src/generated/<tag>/` | `freeze` reverts `SnapshotAlreadyFrozen`. `frozen-snapshots-append-only` still runs after the commit |
| regenerate → freeze ordering | structural in `freeze`: the regeneration is a parameter, run before anything is read or written |
| `forge fmt` before the copy | went, and is not needed: `freeze` copies inside the call, and `forge fmt` afterwards formats the record and the candidate identically |
| `<tag>/` byte-identical to `candidate/` | true by construction — `freeze` writes the bytes it just read out of `candidate`. rain.deploy's `testFreezeCopiesTheRegeneratedRollingSnapshot` asserts it |
| the command must NOT create the tag dir | void — under the standard the command is what creates it |
| `bash -euo pipefail` around the consumer command | kept, in the `run:` step |

No guard is left without a counterpart, so nothing was reimplemented on the rainix side. rain.deploy additionally holds two guards this PR never had: `EmptyRelease` (a release naming no contracts) and `NonMonotonicRelease` (a tag that does not strictly follow every tag already frozen).

## Found while here, not fixed

- "Set foundry.toml version to the release version" rewrites the FIRST `version =` line with `sed`, while `deployTag` reads `.external.package.version` by TOML path. They agree on rain.deploy's `foundry.toml` because that line is first. Pre-existing on `main`, untouched here.
- `README.md` line 150 still documents `rainix-copy-artifacts` as running `./script/BuildPointers.sol`, while the workflow itself requires `script/Build.sol` exactly and errors if `src/generated/` is committed without it. Also pre-existing on `main`.
- The PR title and body above still describe the deleted design.

